PR #4, opened 2021-03-19
Add support for Debian & Ubuntu logos

saloniamatteo:
This pull request adds support for the Debian & Ubuntu logos, using red and white for both logos.

Previews:

debian
ubuntu


saloniamatteo:
Full previews:

debian
ubuntu

PR #9, opened 2021-03-22
Fix GPU value not showing properly

saloniamatteo:
This Pull Request just increases `FASTFETCH_STRBUF_DEFAULT_ALLOC` from `64` to `128`.

It also adds `ctags` & `cscope` files to `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 **/build/
 **/.vscode/
+cscope.*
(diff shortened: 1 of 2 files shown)


LinusDierheimer:
Not merging this as this doesn't fix the issue but just making it more unlikely to appear while doubling ram usage

Issue #8, opened 2021-03-22
GPU value is displayed as "troller"


saloniamatteo:
![gpu](https://user-images.githubusercontent.com/28765699/112031526-5a545000-8b33-11eb-8c22-3fc2cdf9f10e.png)

LinusDierheimer:
Can you please give me the gpu line(s) in the output of lspci?

saloniamatteo:
Removing cached contents with `rm -rf ~/.cache/fastfetch` fixes this problem the next time the program is ran, but on the second execution, the issue comes back.

LinusDierheimer:
So running with --recache gives the right result? Is troller a substring of the gpu name?

saloniamatteo:
Just to be clear, here's a preview:

https://user-images.githubusercontent.com/28765699/112032502-5ecd3880-8b34-11eb-8ee2-fc98928eedf4.mp4

> So running with --recache gives the right result? Is troller a substring of the gpu name?

Yes, as shown in the video.

LinusDierheimer:
Can you cat ~/.cache/fastfetch/GPU0 to look if the problem is saving or loading the cache?

saloniamatteo:
`cat ~/.cache/fastfetch/GPU0` indeed shows "Intel Corporation 4th Gen Core Processor Integrated Graphics Controller", before and after running without recache.

LinusDierheimer:
Thanks, i think the problem is that the string is longer than 64 bytes, which leads to FFstrbuf needing to resizes its capacity. Somehow the original 64 bytes seem to be discarded. I will take a look at this in around a hour. If you're interested the method is ffPrintCachedValue in src/common.c

saloniamatteo:
Thanks, I'll look into it; if I fix the problem, I'll open a PR.

saloniamatteo:
Opened PR #9, closing.

LinusDierheimer:
Found the bug. From https://man7.org/linux/man-pages/man2/read.2.html: 
> On success, the number of bytes read is returned (zero indicates end of file), _and the file position is advanced by this number_.
So we need to rewind the file descriptor each time before we try reading with a larger buffer.
Additional we should call ffStrbufInitA to set a larger buffer in ffPrintCachedValue because i think cached values being larger than 64 bytes is a common thing. I will implement this soon.

saloniamatteo:
[...]

PR #10, opened 2021-03-22
Fix GPU value not showing up correctly

saloniamatteo:
From #9: added `ctags` & `cscope` files to `.gitignore`

New: memory usage is the same (`128` in the previous PR, `64` now).

When reading from fd (file descriptor) with `read()`, rewind fd by 1.
If the return value of `read()` is 0, we have reached the end of the file (`EOF`),
as stated in `man 2 read`:

```
On success, the number of bytes read is returned
(zero indicates end of file),
and the file position is advanced by this number.
```

This PR fixes issue #8 at LinusDierheimer/fastfetch.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 **/build/
 **/.vscode/
+cscope.*
+tags
--- a/src/common.c
+++ b/src/common.c
@@ -239,11 +239,15 @@ void ffGetFileContent(const char* fileName, FFstrbuf* buffer)
         return;
 
     ssize_t readed;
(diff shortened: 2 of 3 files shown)


LinusDierheimer:
This doesn't rewind the file but just makes fd pointing to another position im memory, which may even be another file. I already have implemented the right behaviour and after short testing i will commit. Thanks for the effort

saloniamatteo:
OK. Can you at least add those changes made to `.gitignore` on my behalf?

LinusDierheimer:
This is done.

saloniamatteo:
Thanks.

PR #15, opened 2021-04-21
Fix Void Linux recognition

DarNCelsius:
Added code to strip unescaped qoutes in ID string. Tested on Void and Manjaro.
voidlinux


DarNCelsius:
I don't know if any distro has quotes in their ID though. Might be better just to remove the qoutes?

LinusDierheimer:
See https://github.com/LinusDierheimer/fastfetch/issues/14#issuecomment-825137203. I think this is the better solution because it yields better performance and results.

PR #19, opened 2021-04-24
Added additional Wm Theme support

DarNCelsius:
Support for LXQt, LXDE and Openbox.
wmthemepng

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,7 +19,7 @@ execute_process(
 
 set(PROJECT_VERSION "r${GIT_REV_LIST}.${GIT_REV_PARSE}")
 
-set(CMAKE_C_FLAGS "${CMAKE_C_FLAFS} -Wall -Wextra -O3")
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wextra -O3")
 set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,-O3")
 
 if(BUILD_TESTS)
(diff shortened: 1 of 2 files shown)


LinusDierheimer:
This is good work, but i wont merge it.

First of all this adds a dependency to a dynamic library which is not guaranteed to be present on every Linux system so fastfetch wouldn't be a standalone executable anymore.

Secondly i'm very unhappy with the WM / DE code im currently using and planing to replace it soon. The problem is that im currently relying on `getenv("XDG_SESSION_DESKTOP")` to get the DE, but if somebody uses only a WM (i3 for example) `getenv("XDG_SESSION_DESKTOP")` will return `i3-wm`. On the other side it is possible (and actually being done) to use a different WMs in DEs so for example a `$XDG_SESSION_DESKTOP` of `Plasma` doesn't mean we are using KWin as WM.
Once a good WM / DE detection is written we can think about supporting different WM Themes, but now is not the right time as our internal WM interface might change dramatically. 

Thanks anyway for pointing out the Typo in `CMAKE_C_FLAFS`

DarNCelsius:
This only deals with Openbox as WM, and conditionally loads config files based on DE, or vanilla openbox as default. I don't see any pitfalls there. Worked well on all three.

The implementation seems a bit clunky - some functions could reside elsewhere, like io.c or something I guess.
I also see you have a lot of custom string functions, idk if using them is better.

Is there a way to exclude xml handling if the library is not found on the users system at compile time?
Would it be better with an in-house parser?

LinusDierheimer:
I don't think we need a xml parser to only get one value. Probably a ffParsePropFileHome with a clever regex is enough,
If we really need to load the library we need to do it with dlopen / dlsym / dlclose. 
But again, the project is not yet in a state to add such big specific implementations.

DarNCelsius:
Lol, I was actually trying different regexes before I looked into xml parsing. I could never get decent output though.
What I meant with parser is just something like a variation of ffParsePropFileHome.

You have your first vote on the AUR, btw. :)

PR #21, opened 2021-04-25
Added additional WM Theme support

DarNCelsius:
Supports LXQt, LXDE and Openbox
LXDE:
WMThemeLXDE
LXQt:
WMThemeLXQt
Openbox:
WMThemeOB
My second try, sir. It's faster than the xml version.

### DIFF
--- a/src/modules/wmtheme.c
+++ b/src/modules/wmtheme.c
@@ -1,4 +1,6 @@
 #include "fastfetch.h"
+#include "string.h"
+#include "util/FFstrbuf.h"
 
 #define FF_WMTHEME_MODULE_NAME "WM Theme"
 #define FF_WMTHEME_NUM_FORMAT_ARGS 1
@@ -33,6 +35,79 @@ static void printKWin(FFinstance* instance)
     printWMTheme(instance, theme);
 }
 
+void ffGetOBThemeName(FFinstance* instance, const char* fName, char* buffer)
+{
+    FFstrbuf absolutePath, themeStrbuf;
+    ffStrbufInitA(&absolutePath, 64);
+    ffStrbufAppendS(&absolutePath, instance->state.passwd->pw_dir);
+    ffStrbufAppendS(&absolutePath, fName);
+    ffStrbufInitA(&themeStrbuf, 256);
+    
+    char* line = NULL;
+    size_t len = 0;
+
+    FILE* file = fopen(absolutePath.chars, "r");
+    if(file == NULL)
+        return; // handle errors in higher functions
+        
+    while (getline(&line, &len, file) != -1)
+    {
+        if (strstr(line, "<theme>") != 0)
+            break;
+    }
+    while (getline(&line, &len, file) != -1)
+    {
+        if (strstr(line, "<name>") != 0)
+        {
+            const char* delStrs[] = {"<name>", "</name>"};
+            
+            ffStrbufAppendS(&themeStrbuf, line);
+            ffStrbufRemoveStringsA(&themeStrbuf, 2, delStrs);
+            ffStrbufTrimRight(&themeStrbuf, '\n');
+            ffStrbufTrim(&themeStrbuf, ' ');
+
+            strcpy(buffer, themeStrbuf.chars);
+        }
+        else if (strstr(line, "</theme>") != 0)
+            break;
+        break;
+    }
+        
+    fclose(file);
+    if(line != NULL)
+        free(line);
+
+    ffStrbufDestroy(&absolutePath);
+    ffStrbufDestroy(&themeStrbuf);
+}
+
+static void printOpenbox(FFinstance* instance)
+{
+	char relPath[64];
+	char theme[256];
+	theme[0] = '\0';
+
+	const char* deName = getenv("XDG_SESSION_DESKTOP");
+
+	if (strcmp(deName, (const char*)"LXQt Desktop") == 0)
+		strcpy(relPath, "/.config/openbox/lxqt-rc.xml");
+	else if(strcmp(deName, (const char*)"LXDE") == 0)
+		strcpy(relPath, "/.config/openbox/lxde-rc.xml");
+	else 
+		strcpy(relPath, "/.config/openbox/rc.xml");
+
+	ffGetOBThemeName(instance, relPath, theme);
+
+	if(theme[0] == '\0')
+	{
+		ffPrintError(instance, FF_WMTHEME_MODULE_NAME, 0, &instance->config.wmThemeKey, &instance->config.wmThemeFormat, FF_WMTHEME_NUM_FORMAT_ARGS, "Couldn't find theme name in \"%s\"", relPath);
+		return;
+	}
+
+	printWMTheme(instance, theme);
+}
+
+
 void ffPrintWMTheme(FFinstance* instance)
 {
     const FFWMResult* result = ffCalculateWM(instance);
@@ -45,6 +120,8 @@ void ffPrintWMTheme(FFinstance* instance)
 
     if(ffStrbufIgnCaseCompS(&result->prettyName, "KWin") == 0)
         printKWin(instance);
+    else if(ffStrbufIgnCaseCompS(&result->prettyName, "Openbox") == 0)
+        printOpenbox(instance);
     else
         ffPrintError(instance, FF_WMTHEME_MODULE_NAME, 0, &instance->config.wmThemeKey, &instance->config.wmThemeFormat, FF_WMTHEME_NUM_FORMAT_ARGS, "Unknown WM: %s", result->prettyName.chars);
 }


LinusDierheimer:
This is a very good start, but there are a few issues that should be fixed before merging.

First of all some syntax sugar:  
* your `ffGetOBThemeName` function should be declared `static` as it is used only in this compilation unit. Therefore the `ff` prefix should be removed as this indicates a global function.
* you don't need `#include util/FFstrbuf.h` as this is already done in `fastfetch.h` 
* try to avoid `strcpy` as this function has serious issues when operating on a small buffer. The `ffStrbufAppend*` implementations resize the buffer if necessary Use them if this doesn't mean an unnecessary copy to the Strbuf.

But more important i think the `ffGetOBThemeName` function is redundant. I think something like this should work:
```c
...
ffParsePropFileHome(instance, filePath.chars, "<theme>%s</theme>", theme.chars);
ffStrbufRecalculateLength(&theme); 
...
```

Since the latest commit there is a `ffGetSessionDesktop()` function which should be used over `getenv("XDG_SESSION_DESKTOP")`. You can look at the new DE / WM code to see how it is used.

The behavior and values of DE / WM have changed in the latest commit, as WM is now preferred over DE. You might need to adjust / test your code again, sorry for that.

DarNCelsius:
I'm on it.

DarNCelsius:
`<theme>%s</theme>` does not work. Here's a snippet to show you what the file is like:
```
....
 </placement>
  <theme>
    <name>Matcha-dark-azul</name>
    <titleLayout>NLIMC</titleLayout>
    <!--
      available characters are NDSLIMC, each can occur at most once.
      N: window icon
      L: window label (AKA title).
      I: iconify
      M: maximize
      C: close
      S: shade (roll up/down)
      D: omnipresent (on all desktops).
  -->
    <keepBorder>yes</keepBorder>
....
```
What I did in the commit is scan for the "theme" line, then the subsequent "name" line, and sanitize that. If you can produce a regex to extract this, we'd be golden. :)

LinusDierheimer:
This is almost good, just a few more small things:
1) In this code here:
```c
        ffStrbufTrimRight(&theme, '\n');
        ffStrbufTrim(&theme, ' ');
    }
    else if(strstr(line, "</theme>") != 0) // sanity check
        break;
    break;
}
```
I think we should break after the `ffStrbufTrim(&theme, ' ');`. So we continue immediately with the printing after finding the name. The second break however is too much, as it breaks the loop if the first line in the `<theme> ... </theme>` section does not contain `<name> ... </name>`. XML standard says that the order is not important, so we should just break after finding the closing `</theme>` (like you already implemented). Updated the code would look something like this:
```c
      ffStrbufTrimRight(&theme, '\n');
      ffStrbufTrim(&theme, ' ');
      break;
  } 
 
  if(strstr(line, "</theme>") != 0) // sanity check
      break;
}
```
2) In this code:
```c
FILE* file = fopen(absolutePath.chars, "r");
if(file == NULL)
    return; // handle errors in higher functions
```
It seems like you copied this over from `ffParsePropFile`. However we are in a higher level function, so there should be a `ffPrintError` before the return with a proper error message.
  
3) In this code:
```c
const char* delStrs[] = {"<name>", "</name>"};

ffStrbufAppendS(&theme, line);
ffStrbufRemoveStringsA(&theme, 2, delStrs);
```
 I dont think `ffStrbufRemoveStringsA` is needed as we remove only two strings, so this can very easily be done inline. A `ffStrbufRemoveStrings(&theme, 2, "<name>", "</name>");` should be enough. The array declaration of `delStrs` is redundant with that.

4) In this code:
```c
if(strcmp(deName, (const char*)"LXQt") == 0)
    ffStrbufAppendS(&absolutePath, ".config/openbox/lxqt-rc.xml");
else if(strcmp(deName, (const char*)"LXDE") == 0)
    ffStrbufAppendS(&absolutePath, ".config/openbox/lxde-rc.xml");
```
The `strcmp` calls should be replaced with `strcasecmp` as this way we are on a safer side to always match correctly (Im not sure if env variables are required to be case sensitive).

5) There is nothing wrong, in the following just some syntax sugar and to keep consistency with the rest of the code. If you don't want to change it, you don't have to:
```c
fclose(file);
if(line != NULL) {
    free(line);
}
if(theme.length == 0)
```
* The curly brackets around the first if are are not needed
* There should be a break between the to ifs
* The `fclose(file);` should be after the first if block as it is common and good practice to close / cleanup things in the reversed order as they were opened / created (the `getline` call allocates the space).
* Same thing for `ffStrbufDestroy(&absolutePath); / ffStrbufDestroy(&theme);`

Im sorry if i am annoying you with my critic, i just try to keep code quality high. If you don't understand the reason behind a request (or because of anything else) feel free to ask. I try to give good feedback but i am surely no professional in it.

DarNCelsius:
> This is almost good, just a few more small things:
> 
> 1. In this code here:
> 
> ```c
>         ffStrbufTrimRight(&theme, '\n');
>         ffStrbufTrim(&theme, ' ');
>     }
>     else if(strstr(line, "</theme>") != 0) // sanity check
>         break;
>     break;
> }
> ```
> 
> I think we should break after the `ffStrbufTrim(&theme, ' ');`. So we continue immediately with the printing after finding the name. The second break however is too much, as it breaks the loop if the first line in the `<theme> ... </theme>` section does not contain `<name> ... </name>`. XML standard says that the order is not important, so we should just break after finding the closing `</theme>` (like you already implemented). Updated the code would look something like this:
> 
> ```c
>       ffStrbufTrimRight(&theme, '\n');
>       ffStrbufTrim(&theme, ' ');
>       break;
>   } 
>  
>   if(strstr(line, "</theme>") != 0) // sanity check
>       break;
> }
> ```
Good catch - it sure does fail. I was trying to see if someone botched an edit of the file or something, so we'd reached `</theme>` without finding `<name>`.
> 
> 1. In this code:
> 
> ```c
> FILE* file = fopen(absolutePath.chars, "r");
> if(file == NULL)
>     return; // handle errors in higher functions
> ```
> 
> It seems like you copied this over from `ffParsePropFile`. However we are in a higher level function, so there should be a `ffPrintError` before the return with a proper error message.
> 
I try to copy as much as I can from the code to keep it as similar as possible. This was moved over from the removed function `ffGetOBThemeName`, where it would have worked, so it needs fixing now.
> 1. In this code:
> 
> ```c
> const char* delStrs[] = {"<name>", "</name>"};
> 
> ffStrbufAppendS(&theme, line);
> ffStrbufRemoveStringsA(&theme, 2, delStrs);
> ```
> 
> I dont think `ffStrbufRemoveStringsA` is needed as we remove only two strings, so this can very easily be done inline. A `ffStrbufRemoveStrings(&theme, 2, "<name>", "</name>");` should be enough. The array declaration of `delStrs` is redundant with that.
> 
Nice. I didn't know you could do that. I'm still not sure what all the Strbuf functions do. I was trying to find a `strstr()` equivalent like
```c
bool ffStrbufContains(FFstrbuf* buf, const char* value)
{
    return (strstr(buf->chars, value) != 0);
}
```
> 1. In this code:
> 
> ```c
> if(strcmp(deName, (const char*)"LXQt") == 0)
>     ffStrbufAppendS(&absolutePath, ".config/openbox/lxqt-rc.xml");
> else if(strcmp(deName, (const char*)"LXDE") == 0)
>     ffStrbufAppendS(&absolutePath, ".config/openbox/lxde-rc.xml");
> ```
> 
> The `strcmp` calls should be replaced with `strcasecmp` as this way we are on a safer side to always match correctly (Im not sure if env variables are required to be case sensitive).
> 
Will do.
> 1. There is nothing wrong in the following, just some syntax sugar and to keep consistency with the rest of the code. If you don't want to change it, you don't have to:
> 
> ```c
> fclose(file);
> if(line != NULL) {
>     free(line);
> }
> if(theme.length == 0)
> ```
> 
> * The curly brackets around the first if are are not needed
> * There should be a break between the to ifs
> * The `fclose(file);` should be after the first if block as it is common and good practice to close / cleanup things in the reversed order as they were opened / created (the `getline` call allocates the space).
> * Same thing for `ffStrbufDestroy(&absolutePath); / ffStrbufDestroy(&theme);`
> 
I was adding the curly brackets as it stopped clang from complaining. I'll look into the break and the other stuff.
> Im sorry if i am annoying you with my critic, i just try to keep code quality high. If you don't understand the reason behind a request (or because of anything else) feel free to ask. I try to give good feedback but i am surely no professional in it.

Dont' be sorry - I'm as anal about this as you are, you're just a more proficient coder. :)

LinusDierheimer:
I think you missunderstood point 2, i meant something like this:
```c
FILE* file = fopen(absolutePath.chars, "r");

if(file == NULL)
{
    ffPrintError(...);
    return;
}

while(getline(&line, &len, file) != -1)
```
In the current implementation, if this error occures, it will continue with the `if(theme.length == 0)` check after printing it, which will always be true, and therefor print a seccond error.
This way we also don't need to indent the loop part that much.

Rest is fine :)

LinusDierheimer:
> I think you misunderstood point 2, i meant something like this:
> 
> ```c
> FILE* file = fopen(absolutePath.chars, "r");
> 
> if(file == NULL)
> {
>     ffPrintError(...);
>     return;
> }
> 
> while(getline(&line, &len, file) != -1)
> ```
> 
> In the current implementation, if this error occurs, it will continue with the `if(theme.length == 0)` check after printing it, which will always be true, and therefore print a second error.
> This way we also don't need to indent the loop part that much.
> 
> Rest is fine :)

I just noticed another thing: You must do a `ffStrbufDestroy(&absolutPath)` in this error handling code.

Also the theme strbuf should be declared / allocated after this check to avoid unnecessary ram allocation.

LinusDierheimer:
Looks good, thank you!

PR #32, opened 2021-05-08
Added Terminal Font for Tilix

DarNCelsius:
This adds Terminal Font recognition for Tilix's default profile. I can't find a way to detect which profile is the active one.
tilixTFont

### DIFF
--- a/src/modules/terminalfont.c
+++ b/src/modules/terminalfont.c
@@ -70,9 +70,8 @@ static void printXFCE4Terminal(FFinstance* instance)
 static void printTilixTerminal(FFinstance* instance)
 {
     const char* fontName = NULL;
-    const char* defaultProfile = NULL;
 
-    defaultProfile = ffSettingsGetGsettings(instance, "com.gexperts.Tilix.ProfilesList", NULL, "default", FF_VARIANT_TYPE_STRING).strValue;
+    const char* defaultProfile = ffSettingsGetGsettings(instance, "com.gexperts.Tilix.ProfilesList", NULL, "default", FF_VARIANT_TYPE_STRING).strValue;
 
     if (!defaultProfile)
     {
@@ -83,7 +82,7 @@ static void printTilixTerminal(FFinstance* instance)
     FFstrbuf key;
     ffStrbufInitAS(&key, 64, "/com/gexperts/Tilix/profiles/");
     ffStrbufAppendS(&key, defaultProfile);
-    int keyLen = key.length;
+    uint32_t keyLen = key.length;
     ffStrbufAppendS(&key, "/use-system-font");
 
     FFvariant res = ffSettingsGetDConf(instance,key.chars, FF_VARIANT_TYPE_BOOL);
--- a/src/modules/terminalfont.c
+++ b/src/modules/terminalfont.c
@@ -69,37 +69,33 @@ static void printXFCE4Terminal(FFinstance* instance)
 
 static void printTilixTerminal(FFinstance* instance)
 {
-    static const char* fontName = NULL;
-    static const char* defaultProfile = NULL;
-    static FFstrbuf key;
-    static FFvariant res;
-    res.strValue = NULL;
-    ffStrbufInitAS(&key, 64, "/com/gexperts/Tilix/profiles/");
+    const char* fontName = NULL;
+    const char* defaultProfile = NULL;
 
-    defaultProfile = (ffSettingsGetGsettings(instance, "com.gexperts.Tilix.ProfilesList", NULL, "default", FF_VARIANT_TYPE_STRING)).strValue;
+    defaultProfile = ffSettingsGetGsettings(instance, "com.gexperts.Tilix.ProfilesList", NULL, "default", FF_VARIANT_TYPE_STRING).strValue;
 
     if (!defaultProfile)
     {
         ffPrintError(instance, FF_TERMFONT_MODULE_NAME, 0, &instance->config.termFontKey, &instance->config.termFontFormat, FF_TERMFONT_NUM_FORMAT_ARGS, "Couldn't find \"Default\" profile in Tilix settings");
-        ffStrbufDestroy(&key);
         return;
     }
-    else
+
+    FFstrbuf key;
+    ffStrbufInitAS(&key, 64, "/com/gexperts/Tilix/profiles/");
+    ffStrbufAppendS(&key, defaultProfile);
+    int keyLen = key.length;
+    ffStrbufAppendS(&key, "/use-system-font");
+
+    FFvariant res = ffSettingsGetDConf(instance,key.chars, FF_VARIANT_TYPE_BOOL);
+
+    if(res.boolValueSet && !res.boolValue) // custom font
     {
-        ffStrbufAppendS(&key, defaultProfile);
-        ffStrbufAppendS(&key, "/use-system-font");
-
-        res = ffSettingsGetDConf(instance,key.chars, FF_VARIANT_TYPE_BOOL);
-
-        if(res.boolValueSet && res.boolValue == 0) // custom font
-        {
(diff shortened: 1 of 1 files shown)


src/modules/terminalfont.c
```diff
@@ -69,6 +67,51 @@ static void printXFCE4Terminal(FFinstance* instance)
     printTerminalFont(instance, font);
 }
 
+static void printTilixTerminal(FFinstance* instance)
+{
+    static const char* fontName = NULL;
+    static const char* defaultProfile = NULL;
+    static FFstrbuf key;
+    static FFvariant res;
```
LinusDierheimer, line 75: Why is `static` needed here? They are never needed outside of the function. This way it adds unnecessary symbols.

src/modules/terminalfont.c
```diff
@@ -69,6 +67,51 @@ static void printXFCE4Terminal(FFinstance* instance)
...
+    res.strValue = NULL;
```
LinusDierheimer, line 76: No need to do this explicit, the `ffSettingsGet*` functions do this for you.

src/modules/terminalfont.c
```diff
@@ -69,6 +67,51 @@ static void printXFCE4Terminal(FFinstance* instance)
...
+    ffStrbufInitAS(&key, 64, "/com/gexperts/Tilix/profiles/");
+
+    defaultProfile = (ffSettingsGetGsettings(instance, "com.gexperts.Tilix.ProfilesList", NULL, "default", FF_VARIANT_TYPE_STRING)).strValue;
```
LinusDierheimer, line 79: Why are the braces needed?

src/modules/terminalfont.c
```diff
@@ -69,6 +67,51 @@ static void printXFCE4Terminal(FFinstance* instance)
...
+
+    if (!defaultProfile)
+    {
+        ffPrintError(instance, FF_TERMFONT_MODULE_NAME, 0, &instance->config.termFontKey, &instance->config.termFontFormat, FF_TERMFONT_NUM_FORMAT_ARGS, "Couldn't find \"Default\" profile in Tilix settings");
+        ffStrbufDestroy(&key);
```
LinusDierheimer, line 84: Key is never needed at this point, so we destroy it unused. Wouldn't it make more sense to declare it after the `if`?

src/modules/terminalfont.c
```diff
@@ -69,6 +67,51 @@ static void printXFCE4Terminal(FFinstance* instance)
...
+        return;
+    }
+    else
+    {
+        ffStrbufAppendS(&key, defaultProfile);
+        ffStrbufAppendS(&key, "/use-system-font");
+
+        res = ffSettingsGetDConf(instance,key.chars, FF_VARIANT_TYPE_BOOL);
```
LinusDierheimer, line 92: I think an inline declaration of `res` makes sense here:
```c
FFvariant res = ffSettingsGetDConf(instance,key.chars, FF_VARIANT_TYPE_BOOL);
```

src/modules/terminalfont.c
```diff
@@ -69,6 +67,51 @@ static void printXFCE4Terminal(FFinstance* instance)
...
+
+        if(res.boolValueSet && res.boolValue == 0) // custom font
```
LinusDierheimer, line 94: A  `!res.boolValue` would be more readable

src/modules/terminalfont.c
```diff
@@ -69,6 +67,51 @@ static void printXFCE4Terminal(FFinstance* instance)
...
+        {
+            ffStrbufRemoveStrings(&key, 1, "/use-system-font");
```
LinusDierheimer, line 96: Removing strings is very slow. Maybe store the length of key before `ffStrbufAppendS(&key, "/use-system-font")` and use a `ffSubstrBefore` here.

src/modules/terminalfont.c
```diff
@@ -69,6 +67,51 @@ static void printXFCE4Terminal(FFinstance* instance)
...
+            ffStrbufAppendS(&key, "/font");
+            fontName = (ffSettingsGetDConf(instance, key.chars, FF_VARIANT_TYPE_STRING)).strValue;
```
LinusDierheimer, line 98: Again, why the braces?

src/modules/terminalfont.c
```diff
@@ -69,6 +67,51 @@ static void printXFCE4Terminal(FFinstance* instance)
...
+        }
+        else if(!res.boolValueSet || res.boolValue == 1) // system font
```
LinusDierheimer, line 100: `res.boolValue == 1` is equivalent to a simple `res.boolValue`

src/modules/terminalfont.c
```diff
@@ -69,6 +67,51 @@ static void printXFCE4Terminal(FFinstance* instance)
...
+            fontName = (ffSettingsGetDConf(instance, "/org/gnome/desktop/interface/monospace-font-name", FF_VARIANT_TYPE_STRING)).strValue;
```
LinusDierheimer, line 101: You already know it, right?

LinusDierheimer (changes requested):
Tried the Github file review feature. I hope this way it is even easier for you. Just some small things, overall looking really good.

LinusDierheimer:
Turns out, it shows more lines than marked. Always look at the bottom one to understand what it wrote.

src/modules/terminalfont.c
```diff
@@ -69,6 +67,47 @@ static void printXFCE4Terminal(FFinstance* instance)
     printTerminalFont(instance, font);
 }
 
+static void printTilixTerminal(FFinstance* instance)
+{
+    const char* fontName = NULL;
+    const char* defaultProfile = NULL;
+
+    defaultProfile = ffSettingsGetGsettings(instance, "com.gexperts.Tilix.ProfilesList", NULL, "default", FF_VARIANT_TYPE_STRING).strValue;
```
LinusDierheimer, line 75: Why not do this in one line?

src/modules/terminalfont.c
```diff
@@ -69,6 +67,47 @@ static void printXFCE4Terminal(FFinstance* instance)
...
+    if (!defaultProfile)
+    {
+        ffPrintError(instance, FF_TERMFONT_MODULE_NAME, 0, &instance->config.termFontKey, &instance->config.termFontFormat, FF_TERMFONT_NUM_FORMAT_ARGS, "Couldn't find \"Default\" profile in Tilix settings");
+        return;
+    }
+
+    FFstrbuf key;
+    ffStrbufInitAS(&key, 64, "/com/gexperts/Tilix/profiles/");
+    ffStrbufAppendS(&key, defaultProfile);
+    int keyLen = key.length;
```
LinusDierheimer, line 86: `key.length` is a `uint32_t`. We should keep this to prevent casting errors.

LinusDierheimer (commented):
We are almost there!

src/modules/terminalfont.c
DarNCelsius: Wonder why there is no warning about stuff like this?

LinusDierheimer: I didn't enable -Wconversion on the standard build, because there are situations there are situations where it is needed to do a conversion that the c standard doesn't like.
For example, ISO c forbids a conversion from a void pointer to a function pointer, however POSIX requires us to do exactly this in functions like [dlysm](https://linux.die.net/man/3/dlsym)...
To build with all warnings, run `tests/performance.sh`, but don't be shocked by all the warnings you will get, most of them are allowed by POSIX and just not by ISO c, and therefore allowed in a linux only context.
I decided to do this, because i don't want users to see warnings when building the project (in example from the AUR).

LinusDierheimer (approved):

LinusDierheimer:
Happy with this one, thank you!

PR #42, opened 2021-05-15
Typos and grammar

DarNCelsius:
I only did comments and help text - You might wanna change the "--seperate" option.
```
sep·a·rate
    a. To set, force, or keep apart: The referee separated the two boxers.
    b. To put space between; space apart or scatter: small farms that were separated one from another by miles of open land.
    c. To form a border or barrier between (two areas or groups): A hedge separates the two yards.
    d. To place in different groups; sort: separate mail by postal zones.
```


LinusDierheimer:
Woah thank you very much for those! Feel free to open a PR for the option.

PR #44, opened 2021-05-15
Added branch detection for Manjaro

DarNCelsius:
I realize you said it wasn't very useful, which I agree with for Arch, but I think it makes sense on Manjaro.
Arch is set up differently (I think), whereas Manjaro only has one active branch at any time.
branchdetect
Some people do change branches when testing, or if they simply want a file that is available only in the unstable repo.
Certainly very useful for people that don't always remember to go back to stable. :wink:

### DIFF
--- a/src/modules/packages.c
+++ b/src/modules/packages.c
@@ -57,7 +57,7 @@ static void getActiveBranch()
 {
     FILE* file = fopen("/etc/pacman-mirrors.conf", "r");
     if(file == NULL)
-        return; // No file or not Manjaro/based
+        return; // No file or not Manjaro[based]
 
     char* line = NULL;
     size_t len = 0, strip;
@@ -75,10 +75,7 @@ static void getActiveBranch()
 
     fclose(file);
 
-    if(match > 0)
-        printf("[%s]", line);
-    else
-        printf("[stable]"); // defaults to stable
+    match > 0 ? printf("[%s]", line) : printf("[stable]");
 
     if(line != NULL)
         free(line);
--- a/src/modules/packages.c
+++ b/src/modules/packages.c
@@ -53,12 +53,6 @@ static uint32_t getNumStrings(const char* filename, const char* needle)
     return count;
 }
 
-static void getActiveBranchManjaro(FFstrbuf buffer)
-{
-    if(ffParsePropFile("/etc/pacman-mirrors.conf", "Branch = ", &buffer) && buffer.length == 0)
-        ffStrbufSetS(&buffer, "stable");
-}
-
 void ffPrintPackages(FFinstance* instance)
 {
     uint32_t pacman = getNumElements("/var/lib/pacman/local", DT_DIR);
@@ -76,8 +70,9 @@ void ffPrintPackages(FFinstance* instance)
     }
     FFstrbuf manjaroBranch;
     ffStrbufInit(&manjaroBranch);
-    getActiveBranchManjaro(manjaroBranch);
-    ffStrbufRecalculateLength(&manjaroBranch); // needed for later length comparison...
+
+    if(ffParsePropFile("/etc/pacman-mirrors.conf", "Branch =", &manjaroBranch) && manjaroBranch.length == 0)
+        ffStrbufSetS(&manjaroBranch, "stable");
 
     if(instance->config.packagesFormat.length == 0)
     {
@@ -99,6 +94,9 @@ void ffPrintPackages(FFinstance* instance)
             if((all = all - pacman) > 0)
                 printf(", ");
         };
+
+        ffStrbufDestroy(&manjaroBranch);
+
         FF_PRINT_PACKAGE(xbps)
         FF_PRINT_PACKAGE(dpkg)
         FF_PRINT_PACKAGE(flatpak)
@@ -122,5 +120,4 @@ void ffPrintPackages(FFinstance* instance)
             {FF_FORMAT_ARG_TYPE_UINT, &snap}
         });
     }
-    ffStrbufDestroy(&manjaroBranch);
 }
--- a/src/modules/packages.c
+++ b/src/modules/packages.c
@@ -53,6 +53,37 @@ static uint32_t getNumStrings(const char* filename, const char* needle)
     return count;
 }
 
+static void getActiveBranch()
+{
+    FILE* file = fopen("/etc/pacman-mirrors.conf", "r");
+    if(file == NULL)
+        return; // No file or not Manjaro/based
+
+    char* line = NULL;
+    size_t len = 0, strip;
+    uint_fast8_t match = 0;
+
+    while (getline(&line, &len, file) != EOF)
+    {
+        if((strip = strspn(line, "\t ") > 0)) {
+            if(strip != len)
+                memmove(line, line + strip, len + 1 - strip);
+        }
+        if((match = sscanf(line, "Branch = %s", line)) > 0)
+            break;
+    }
+
+    fclose(file);
+
+    if(match > 0)
+        printf("[%s]", line);
+    else
+        printf("[stable]"); // defaults to stable
+
+    if(line != NULL)
+        free(line);
+}
+
 void ffPrintPackages(FFinstance* instance)
 {
     uint32_t pacman = getNumElements("/var/lib/pacman/local", DT_DIR);
@@ -77,6 +108,8 @@ void ffPrintPackages(FFinstance* instance)
         if(name > 0) \
         { \
             printf("%u ("#name")", name); \
+            if(name == pacman)  \
+                getActiveBranch(); \
             if((all = all - name) > 0) \
                 printf(", "); \
         };
--- a/src/modules/packages.c
+++ b/src/modules/packages.c
@@ -53,32 +53,10 @@ static uint32_t getNumStrings(const char* filename, const char* needle)
     return count;
 }
 
-static void getActiveBranch()
+static void getActiveBranchManjaro(FFstrbuf buffer)
 {
-    FILE* file = fopen("/etc/pacman-mirrors.conf", "r");
-    if(file == NULL)
-        return; // No file or not Manjaro[based]
-
-    char* line = NULL;
-    size_t len = 0, strip;
-    uint_fast8_t match = 0;
-
-    while (getline(&line, &len, file) != EOF)
-    {
-        if((strip = strspn(line, "\t ") > 0)) {
-            if(strip != len)
-                memmove(line, line + strip, len + 1 - strip);
-        }
-        if((match = sscanf(line, "Branch = %s", line)) > 0)
-            break;
-    }
-
-    fclose(file);
-
-    match > 0 ? printf("[%s]", line) : printf("[stable]");
-
-    if(line != NULL)
-        free(line);
+    if(ffParsePropFile("/etc/pacman-mirrors.conf", "Branch = ", &buffer) && buffer.length == 0)
+        ffStrbufSetS(&buffer, "stable");
 }
 
 void ffPrintPackages(FFinstance* instance)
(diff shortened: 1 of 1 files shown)


src/modules/packages.c
```diff
@@ -53,6 +53,34 @@ static uint32_t getNumStrings(const char* filename, const char* needle)
     return count;
 }
 
+static void getActiveBranch()
```
LinusDierheimer, line 56: The method name should contain Manjaro to make clear it is a manjaro only function.

src/modules/packages.c
```diff
@@ -53,6 +53,34 @@ static uint32_t getNumStrings(const char* filename, const char* needle)
...
+{
+    FILE* file = fopen("/etc/pacman-mirrors.conf", "r");
+    if(file == NULL)
+        return; // No file or not Manjaro[based]
+
+    char* line = NULL;
+    size_t len = 0, strip;
+    uint_fast8_t match = 0;
```
LinusDierheimer, line 64: Wouldn't it be better if match is the return type of sscanf?

src/modules/packages.c
```diff
@@ -77,6 +105,8 @@ void ffPrintPackages(FFinstance* instance)
         if(name > 0) \
         { \
             printf("%u ("#name")", name); \
+            if(name == pacman)  \
+                getActiveBranch(); \
```
LinusDierheimer, line 109: This seems unsafe, what happens if a package manager has the same amount of packages as pacman? Just don't use the macro for pacman and write it by hand.

src/modules/packages.c
```diff
@@ -53,6 +53,34 @@ static uint32_t getNumStrings(const char* filename, const char* needle)
...
+    uint_fast8_t match = 0;
+
+    while (getline(&line, &len, file) != EOF)
+    {
+        if((strip = strspn(line, "\t ") > 0)) {
+            if(strip != len)
+                memmove(line, line + strip, len + 1 - strip);
+        }
+        if((match = sscanf(line, "Branch = %s", line)) > 0)
+            break;
```
LinusDierheimer, line 73: There is a ffGetPropValue for exactly this. It ignores tabs and spaces at the beginn of the line so it should work flawless.

LinusDierheimer (commented):
I didn't know Manjaro works this way, on Arch you pull from stable / extra / community / etc simultaneously. This way i agree that it might be usefull. 
You should also add a format arg containing the branch.
Since i don't have /etc/pacman-mirrors.conf i don't know how it looks like and can't judge if your code implements it a good way, i simply trust you with this one.
As a tip, i think the simples thing is to create a FFstrbuf after
```c
if(all == 0)
{
    ffPrintError(instance, FF_PACKAGES_MODULE_NAME, 0, &instance->config.packagesKey, &instance->config.packagesFormat, FF_PACKAGES_NUM_FORMAT_ARGS, "No packages from known package managers found");
    return;
}
```

Your function should take it as an argument and pass it to `ffGetPropValue`. So your function always gets called, but only one time per run.
This way it should also be very simple to pass it as a format arg and print it when not using a format string.

src/modules/packages.c
DarNCelsius: Probably - I was looking for the smallest size I could use.

src/modules/packages.c
DarNCelsius: Oops. I was seeing  'pacman' as an enum I could use to identify it.

src/modules/packages.c
DarNCelsius: Ah. That's a recent addition I think. I forgot about that.

DarNCelsius:
Here's pacman-mirrors.conf in its default state:
```
##                                                                                                                                                                                                                                                                                        
## /etc/pacman-mirrors.conf                                                                                                                                                                                                                                                               
##                                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                          
## Branch Pacman should use (stable, testing, unstable)                                                                                                                                                                                                                                   
# Branch = stable                                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                          
## Generation method                                                                                                                                                                                                                                                                      
## 1) rank   - rank mirrors depending on their access time                                                                                                                                                                                                                                
## 2) random - randomly generate the output mirrorlist                                                                                                                                                                                                                                    
# Method = rank                                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                          
## Filename to use when ranking mirrors                                                                                                                                                                                                                                                   
## The file must be present in core repo                                                                                                                                                                                                                                                  
# TestFile = core.db.tar.gz                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                          
## Define protocols and priority                                                                                                                                                                                                                                                          
##   separated by comma 'https,http' or 'http,https'                                                                                                                                                                                                                                      
## ATM available protocols are: http, https, ftp                                                                                                                                                                                                                                          
## Not specifying a protocol will ban the protocol from being used                                                                                                                                                                                                                        
## If a mirror has more than one protocol defined only the first is written to the mirrorlist                                                                                                                                                                                             
## Empty means all in reversed alphabetic order                                                                                                                                                                                                                                           
# Protocols =                                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                          
## When set to False - all certificates are accepted.                                                                                                                                                                                                                                     
## Use only if you fully trust all ssl-enabled mirrors.                                                                                                                                                                                                                                   
# SSLVerify = True                                                                                                                                                                                                                                                                        

```
If the Branch line is like this, it simply defaults to stable.

LinusDierheimer:
I think it can be done very simply then:
```c
FFstrbuf manjaroBranch;
ffStrbufInit(&manjaroBranch);

if(ffParsePropFile("/etc/pacman-mirrors.conf", "Branch =", &manjaroBranch) && manjaroBranch.length == 0)
    ffStrbufSetS(&manjaroBranch, "stable");
```
Remember to pull the commit i just pushed to have the result trimmed with spaces and the second check work.

src/modules/packages.c
DarNCelsius: How do I get the output in front of the comma without modifying the macro?

LinusDierheimer: Just replace `FF_PRINT_PACKAGE(pacman)` with
```c
if(pacman > 0)
{
    printf("%u (pacman)", pacman);
    if(manjaroBranch.length > 0)
        printf("[%s]", manjaroBranch.chars);
    if((all = all - pacman) > 0)
        printf(", ");
};
```

src/modules/packages.c
```diff
@@ -53,6 +53,12 @@ static uint32_t getNumStrings(const char* filename, const char* needle)
     return count;
 }
 
+static void getActiveBranchManjaro(FFstrbuf buffer)
+{
+    if(ffParsePropFile("/etc/pacman-mirrors.conf", "Branch = ", &buffer) && buffer.length == 0)
+        ffStrbufSetS(&buffer, "stable");
+}
+
```
LinusDierheimer, line 61: I don't think we need an extra method just for this.

src/modules/packages.c
```diff
@@ -68,6 +74,10 @@ void ffPrintPackages(FFinstance* instance)
         ffPrintError(instance, FF_PACKAGES_MODULE_NAME, 0, &instance->config.packagesKey, &instance->config.packagesFormat, FF_PACKAGES_NUM_FORMAT_ARGS, "No packages from known package managers found");
         return;
     }
+    FFstrbuf manjaroBranch;
+    ffStrbufInit(&manjaroBranch);
+    getActiveBranchManjaro(manjaroBranch);
+    ffStrbufRecalculateLength(&manjaroBranch); // needed for later length comparison...
```
LinusDierheimer, line 80: This is weird, you should't need it

DarNCelsius:
I have no idea how the format arg stuff works. Is that necessary?

LinusDierheimer:
Its not a must but is usefull. I will write it as soon we merged.

LinusDierheimer (commented):
Just two small things.

src/modules/packages.c
DarNCelsius: What is ?

LinusDierheimer: `ffStrbufRecalculateLength`

DarNCelsius: The strbuf and placement is from your earlier suggestion, the Recalculate was needed to get length comparison to work.

LinusDierheimer: Yea i understood this, the thing is the length should always be up-to-date.

DarNCelsius: In debug, it showed `manjaroBranch.length` to 0. If you want to test it, just create the file and use the contents I provided earlier.

LinusDierheimer: If i use the default file you provided, the length is expected to be 0. This happens when  `Branch =` was not found, for example commented out. This is why i set the default value when the file was found, but not the property.

LinusDierheimer: Tested it, works perfectly.

DarNCelsius: Length surely shouldn't be 0 when the buffer is set with `ffStrbufSetS(&manjaroBranch, "stable");` ?

LinusDierheimer: That is right, and it is not 0 when i test it. Try with a breakpoint a bit later in the code.

DarNCelsius: Removing that line makes the branch info disappear. IDK what is happening.

LinusDierheimer: Compiler optimizations probably as i force -O3 in CMakeLists.txt. What does a printf say about the length?

DarNCelsius:
```c
        if(pacman > 0)
        {
            printf("%u (pacman)", pacman);
            printf("Len: %d", manjaroBranch.length);
            if(manjaroBranch.length > 0)
                printf("[%s]", manjaroBranch.chars);
            if((all = all - pacman) > 0)
                printf(", ");
        };
```
`Packages: 1499 (pacman)Len: 0, 1 (flatpak)`

LinusDierheimer: Try with %u

DarNCelsius: same

DarNCelsius: Did you rewrite anything at all before testing?

LinusDierheimer: Did some changes to ffGetPropValue, but they should't affect you. They are already pushed so you can just pull them.

DarNCelsius: Pulled 3 hours ago

LinusDierheimer: I've tested with the file not existing, it existing with branch as a comment and with a changed branch. Works all time:
```c
FFstrbuf manjaroBranch;
ffStrbufInit(&manjaroBranch);

if(ffParsePropFile("/etc/pacman-mirrors.conf", "Branch =", &manjaroBranch) && manjaroBranch.length == 0)
    ffStrbufSetS(&manjaroBranch, "stable");

if(instance->config.packagesFormat.length == 0)
{
    ffPrintLogoAndKey(instance, FF_PACKAGES_MODULE_NAME, 0, &instance->config.batteryKey);

    #define FF_PRINT_PACKAGE(name) \
    if(name > 0) \
    { \
        printf("%u ("#name")", name); \
        if((all = all - name) > 0) \
            printf(", "); \
    };

    if(pacman > 0)
    {
        printf("%u (pacman)", pacman);
        if(manjaroBranch.length > 0)
            printf("[%s]", manjaroBranch.chars);
        if((all = all - pacman) > 0)
            printf(", ");
    };
```

DarNCelsius: You did rewrite it...

LinusDierheimer: I just optimized the ffStrbufTrim, like i said it should't change how the method works.

DarNCelsius: Sure enough, now it works. Length was somehow lost using the `getActiveBranchManjaro()` function.
Should it not persist?

LinusDierheimer: Ah now i see your error, you have to pass FFstrbuf as a pointer to persist length. I didn't look at this because i just commented that the function is not needed. Only the data of a FFstrbuf is stored on ram.

DarNCelsius: Well I did say "anything at all", Linus. ;)

LinusDierheimer: My fault :)

PR #45, opened 2021-05-15
Tweaks

DarNCelsius:
It's an OCD thing. :grinning:


LinusDierheimer:
I don't see why but also why not so i merge it.

PR #46, opened 2021-05-15
Fixed removeStrings() divider

DarNCelsius:
remstrs3

### DIFF
--- a/src/modules/cpu.c
+++ b/src/modules/cpu.c
@@ -105,7 +105,7 @@ void ffPrintCPU(FFinstance* instance)
         " 2-Core", " 4-Core", " 6-Core", " 8-Core", " 10-Core", " 12-Core", " 14-Core", " 16-Core"
     };
 
-    ffStrbufRemoveStringsA(&namePretty, sizeof(removeStrings) / sizeof(removeStrings[0]), removeStrings);
+    ffStrbufRemoveStringsA(&namePretty, sizeof(removeStrings) / (sizeof(removeStrings) / sizeof(removeStrings[0])), removeStrings);
     ffStrbufSubstrBeforeFirstC(&namePretty, '@'); //Cut the speed output in the name as we append our own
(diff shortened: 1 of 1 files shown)


LinusDierheimer:
This doesn't seem right at all, because `168 / (168 / 8)` is `8` while `168 / 8` is the correct amount of `21`. Does hard coding the this value work?

DarNCelsius:
Hang on...

DarNCelsius:
It alternates with different values. Removes 8-Core, but not Processor and vice versa.

LinusDierheimer:
So the implementation is flawed. Time to look into it.

DarNCelsius:
Using `ffStrbufRemoveStrings(&namePretty, 2, " Processor", " 8-Core");` fails - '8-core' is removed.
Using `ffStrbufRemoveStrings(&namePretty, 1, " Processor");` works.

DarNCelsius:
All these functions end up in  `strbufRemoveTest()`. Is there a condition missing or something in there?

DarNCelsius:
Manipulating `strbuf->length -= k;` with for example `strbuf->length -= k + 1;` on line 316, will change the output of the "Processor" string, so it's definitely passed both checks in the for loop.

DarNCelsius:
Does the order in which the string elements are encountered matter?

Rearranging the array to this:
```c
    const char* removeStrings[] = {
        "(R)", "(r)", "(TM)", "(tm)",
        " Dual-Core", " Quad-Core", " Six-Core", " Eight-Core", " Ten-Core",
        " 2-Core", " 4-Core", " 6-Core", " 8-Core", " 10-Core", " 12-Core", " 14-Core", " 16-Core",
        " CPU", " FPU", " APU", " Processor"
    };
```
fixes it completely for me. This array order mimics the order in the original string - `AMD Ryzen 7 3800X 8-Core Processor`.
remstrs4
Both `8-Core` and `Processor` is removed. Not a proper fix, but should help track it down.

DarNCelsius:
I opened an issue for this #47.

PR #49, opened 2021-05-16
Enhanced LXQt DE Support

DarNCelsius:
Couldn't find version info in any files, so this is as slow as it gets...
lxqtversion
It fetches 3 versions
```
lxqt-session 0.17.1
liblxqt   0.17.0
Qt        5.15.2
```


LinusDierheimer:
Somehow github is broken so i have to comment this way.

In code:
```c
ffStrbufSubstrAfter(&result->deVersion, result->deProcessName.length);
ffStrbufSubstrBefore(&result->deVersion, 6); //X.XX.X
```

This is unsave. I just installed LXQT for testing the PR and i get the following output:
```
"Icon Theme not set. Fallbacking to Oxygen, if installed"
lxqt-session 0.17.1
liblxqt   0.17.0
Qt        5.15.2
```
You will need to to some more advanced parsing. Maybe loop through it like i do here: https://github.com/LinusDierheimer/fastfetch/blob/8af1cde10b2be86c486efcb38ed2ab43a9302b10/src/modules/disk.c#L118 or here: https://github.com/LinusDierheimer/fastfetch/blob/8af1cde10b2be86c486efcb38ed2ab43a9302b10/src/common/init.c#L27 just use '\n' as the seperator and not ':'. You can than use `ffGetPropValue` on each line.

It installed a lot of files with the version 0.17.0 on my system. `/usr/share/cmake/lxqt/lxqt-*.cmake` look promising. Maybe we should fallback to that if slow operations are not allowed.

DarNCelsius:
Hm. Didn't know --version outputs even more garbage....

Yes, I looked at the cmake files after this commit. If they are there for everyone, we should probably use those. The main LXQt About dialog also shows version as 0.17.0.

lxqt-config.cmake: `set(LXQT_VERSION            "0.17.0")`

I'll rewrite it.

DarNCelsius:
Your latest commits made this very easy. :+1: 
I went with checking the `liblxqt` version, as that seems to be more accurate the `lxqt-session`.

src/common/detectWMDE.c
```diff
@@ -275,6 +278,28 @@ static void getXFCE4(FFinstance* instance, FFWMDEResult* result)
...
+        ffProcessAppendStdOut(&result->deVersion, (char* const[]){
+            "lxqt-session",
+            "-v",
+            NULL
+        });
+
+        FF_STRBUF_CREATE(tmp);
+        ffGetPropValueFromLines(result->deVersion.chars , "liblxqt", &tmp);
+        ffStrbufSet(&result->deVersion, &tmp);
+        ffStrbufDestroy(&tmp);
```
LinusDierheimer, line 299: This is a unnecesarry buffer allocation. Since we will always copy back to front, in an index increasing order, we can safely use only one buffer. We set the length to 0, so appending will start at index 0, but leave the null byte at the old index, because `ffgetPropValue*` functions use it to detect the length. Implementation would look something like this:
```c
result->deVersion.length = 0; //don't set '\0' byte
ffGetPropValueFromLines(result->deVersion.chars , "liblxqt", &result->deVersion);
```
I didn't test it, but it _should_ work.

src/common/detectWMDE.c
```diff
@@ -275,6 +278,28 @@ static void getXFCE4(FFinstance* instance, FFWMDEResult* result)
     }
 }
 
+static void getLXQt(FFinstance* instance, FFWMDEResult* result)
+{
+    ffStrbufSetS(&result->deProcessName, "lxqt-session");
+    ffStrbufSetS(&result->dePrettyName, "LXQt");
+    ffParsePropFile("/usr/share/cmake/lxqt/lxqt-config.cmake", "set(LXQT_VERSION", &result->deVersion);
```
LinusDierheimer, line 285: As far as i know, cmake allows spaces, so we should too:
```c
"set ( LXQT_VERSION"
```
A space in the format string matches _any_ amount of spaces / tabs in the scanned string, even none.

src/common/detectWMDE.c
```diff
@@ -275,6 +278,28 @@ static void getXFCE4(FFinstance* instance, FFWMDEResult* result)
...
+
+    if(result->deVersion.length == 0 && instance->config.allowSlowOperations)
```
LinusDierheimer, line 287: Maybe we should try parsing other files before falling back to this. The other CMake file and the pkgconfig file look promising.

src/common/detectWMDE.c
DarNCelsius: It worked. :+1: 
I tried it before without the length adjustment, which gave me all the lines, so I used a tmp buffer instead.

LinusDierheimer: Yes this is poor method naming on my side. `ffGetPropValue*` functions actually append the buffer instead of setting it.

PR #51, opened 2021-05-17
Refactored terminal font functions

DarNCelsius:
Joined Tilix and Gnome terminal font functions, as they were pretty much identical.


DarNCelsius:
I discovered what was happening with GSettings and DConf. The path usage was wrong, so GSettings was pulling from the schema itself, and never the actual setting.

PR #52, opened 2021-05-18
Added Gnome Terminal Fonts

DarNCelsius:
This will compile the project a lot faster:
```sh
cmake --build . "-j$(($(nproc)+1))"

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ KDE Plasma, Gnome, Cinnamon, Mate, XFCE4, LXQt
 
 ##### Terminal fonts:
 ```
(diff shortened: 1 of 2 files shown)


DarNCelsius:
Looks a bit messy. I should probably have used two pull requests.

LinusDierheimer:
About the multithreaded building: I think each user should decide if he wants to do this. There are actually people who prefer to don't hammer there cpu. If you want multithreaded building on your system, use the `MAKEFLAGS` environment variable. Like you already probably use in your /etc/makepkg.conf.

For the README.md update open a new pull request.

PR #53, opened 2021-05-18
fix --help *-format default message

Euro20179:
I couldn't fix wm-theme-format, theme-format, and icon-format

also im not sure but i dont think by default uptime displays the amount of days, but in the help it says it does, correct me if i'm wrong.

### DIFF
--- a/src/fastfetch.c
+++ b/src/fastfetch.c
@@ -269,7 +269,7 @@ static inline void printCommandHelp(const char* command)
     }
     else if(strcasecmp(command, "uptime-format") == 0)
     {
-        constructAndPrintCommandHelpFormat("uptime", "{} days {} hours {} mins", 4,
+        constructAndPrintCommandHelpFormat("uptime", "{1} days {2} hours {3} mins", 4,
             "Days",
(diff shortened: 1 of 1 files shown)


LinusDierheimer:
The formats you "corrected" are actually right. An empty brace pair has the value of the amount it was parsed until occurrence. So the first `{}` is equal to `{1}`, the second `{}` is equal to `{2}` and so on. Maybe i should write some clearer help on this.
 
Uptime actually does print days, but only if your uptime is greater than one day. Like it also only prints hours if your uptime is greater than one hour. Or it prints seconds, but only if, your uptime hasn't reached one minute yet. 
If you wanted to duplicated default uptime with a format string, this would come pretty close:
```
{?1}{1} days, {?}{?2}{2} hours, {?}{?3}{3} mins{?}{/3}{4} seconds{/}
```
However, i purposely decided against using such formats as default in help, because i think they are too complicated to some, especially new, users. This is why i write "something like". Some behavior can't even be reproduced (yet), for example the plural s on days, hours, mins only being present if the value is greater than one.

PR #54, opened 2021-05-19
Update README.md

DarNCelsius:
Added Gnome Terminal to Terminal Fonts section


LinusDierheimer:
Two spaces at the end of a line are actually part of the markdown syntax, they indicate a line break. Otherwise it wouldn't work because it is compiled to HTML. So most of the whites paces are actually intended. More info here: https://www.markdownguide.org/basic-syntax/#line-breaks

DarNCelsius:
Oh wow... I had no idea. I just turned a setting on in Code to remove trailing whitespace on save.

PR #57, opened 2021-05-26
Custom logos read from files

alborrajo:
If the logo specified with ```--logo``` isn't recognized, the program will check if it's a file that can be read, and if it is, the contents of the file will be used. Otherwise it'll use the unknown logo.


README.md
```diff
@@ -2,8 +2,8 @@
 
 fastfetch is a [neofetch](https://github.com/dylanaraps/neofetch) like tool for fetching system information and displaying them in a pretty way. It is written in c to achieve much better performance, in return only linux is supported. It also uses mechanisms like multithreading and caching to finish as fast as possible.  
   
-The output on my personal machine with default configurations looks like this:
-<img src="screenshots/example.png" alt="example output" width="100%"/>  
+This fork adds custom logos read from files:
+<img src="screenshots/amogus.png" alt="example output" width="100%"/>
```
LinusDierheimer, line 6: This changes are fine for your fork but shouldn't be in this repository.

src/common/logo.c
```diff
@@ -249,6 +269,37 @@ static void loadGarudaLogo(FFlogo* logo, bool doColor)
     sprintf(logo->chars[19], FASTFETCH_TEXT_MODIFIER_BOLT"%s                                              "FASTFETCH_TEXT_MODIFIER_RESET, color);
 }
 
+static void loadLogoFromFile(FFlogo* logo, bool doColor, char* pathToFile)
+{
+    strcpy(logo->name, "amogus");
+
+    const char* color = doColor ? "\033[91m" : "";
+    strcpy(logo->color, "\033[91m");
```
LinusDierheimer, line 277: I don't think we should set the same color for every logo loaded from file. Use the `--color` option to set it.

src/common/logo.c
```diff
@@ -249,6 +269,37 @@ static void loadGarudaLogo(FFlogo* logo, bool doColor)
     sprintf(logo->chars[19], FASTFETCH_TEXT_MODIFIER_BOLT"%s                                              "FASTFETCH_TEXT_MODIFIER_RESET, color);
 }
 
+static void loadLogoFromFile(FFlogo* logo, bool doColor, char* pathToFile)
+{
+    strcpy(logo->name, "amogus");
```
LinusDierheimer, line 274: No need to do this, as we never load the logo by name. Maybe just set a null byte at index 0 to be save.

src/common/logo.c
```diff
@@ -249,6 +269,37 @@ static void loadGarudaLogo(FFlogo* logo, bool doColor)
...
+        fgets(buff, sizeof(buff), fp);
+        normalizedBuff = strtok(buff, "\n");
+        
+        sprintf(logo->chars[logoHeight++],   FASTFETCH_TEXT_MODIFIER_BOLT"%s%s"FASTFETCH_TEXT_MODIFIER_RESET, color, normalizedBuff);
+
+        len = utf8_strlen(normalizedBuff);
+        logoWidth = len > logoWidth
+            ? len
+            : logoWidth;
+    }
```
LinusDierheimer, line 295: Wouldn't it be simpler to use `getline`?

src/common/logo.c
```diff
@@ -272,6 +323,8 @@ void ffLoadLogoSet(FFconfig* config, const char* logo)
         loadGarudaLogo(&config->logo, config->colorLogo);
     else if(strcasecmp(logo, "unknown") == 0)
         loadUnknownLogo(&config->logo);
+    else if(access(logo, R_OK) == 0)
+        loadLogoFromFile(&config->logo, config->colorLogo, logo);
```
LinusDierheimer, line 327: I think we should also introduce an extra option for this. Maybe `--logo-file`? This way the user has better control about what logo is loaded. Anyway we can leave this a a fallback for `--logo`.

src/common/logo.c
```diff
@@ -249,6 +269,37 @@ static void loadGarudaLogo(FFlogo* logo, bool doColor)
...
+    int len;
+
+    int logoWidth = 0, logoHeight = 0;
+    while(!feof(fp)) {
+        fgets(buff, sizeof(buff), fp);
+        normalizedBuff = strtok(buff, "\n");
+        
+        sprintf(logo->chars[logoHeight++],   FASTFETCH_TEXT_MODIFIER_BOLT"%s%s"FASTFETCH_TEXT_MODIFIER_RESET, color, normalizedBuff);
+
+        len = utf8_strlen(normalizedBuff);
```
LinusDierheimer, line 291: Im not into this topic, why isn't a simple `strlen` enough? The point of Ascii logos is that they can be encoded using ascii so i'm not sure if we need to support unicode.

alborrajo:
I'll fix all of this when I get some time later. Sorry if some things don't
make much sense, I'm pretty much inexperienced in both C development and
doing pull request.
The logo name being amogus was something I forgot to remove.
strlen gave me wrong results when trying to load logos that included UTF-8
characters, so I added that function to support them.
I didn't know about getline, as I said I barely have experience with C.
I'll take a look at it.

src/common/logo.c
alborrajo: I'm having trouble getting color to work even when using the examples in ```fastfetch --help color```. None of them seem to work, it just prints the color encoding sequence as it is, even on the information lines (OS, host, kernel, etc)

src/common/logo.c
alborrajo: I don't see how using getline would make it simpler, getline gets the same data as fgets, I still have to strip the newline character from the read data

src/common/logo.c
LinusDierheimer: This actually was a bug, fixed in the latest commit.

src/common/logo.c
LinusDierheimer: `getline` returns the number of bytes read, so removing the 'ņ' is faster.

src/common/logo.c
LinusDierheimer: Just like all flags are handled: The last one is used. So you can overwrite previous options.

src/common/logo.c
alborrajo: It's not necessary, but it's a nice bonus. If you don't think there has to be support for UTF8 i can just replace it with strlen

completions/bash
```diff
@@ -192,6 +192,7 @@ __fastfetch_completion()
     local FF_OPTIONS_LOGO=(
         "-l"
         "--logo"
+        "--logo-file"
```
LinusDierheimer, line 195: This is a path option, not a logo option. Logo options are completed with the names of the build in logos.

src/common/logo.c
```diff
@@ -249,29 +269,71 @@ static void loadGarudaLogo(FFlogo* logo, bool doColor)
...
+            buff[lineBytes-1] = '\0';
+        
+        sprintf(logo->chars[logoHeight++],   FASTFETCH_TEXT_MODIFIER_BOLT"%s%s"FASTFETCH_TEXT_MODIFIER_RESET, color, buff);
+
+        lineLength = utf8_strlen(buff);
+        logoWidth = lineLength > logoWidth
+            ? lineLength
+            : logoWidth;
+    }
+
```
LinusDierheimer, line 296: insert a 
```c
if(buff != NULL)
    free(buff);
```
here

LinusDierheimer:
Thanks for the contribution!

LinusDierheimer:
It doesn't load the OS logo anymore on my machine, but i will fix that by my self. You did great work providing this code to build upon.

alborrajo:
Yes, in my machine both work as it should

Issue #67, opened 2021-06-20
Strange KWin WMTheme output


DarNCelsius:
![wmthemekwin](https://user-images.githubusercontent.com/82701459/122669580-5da4b980-d1be-11eb-9e1e-ca8596e6fdd7.png)

LinusDierheimer:
Content of `$XDG_CONFIG_HOME/kwinrc`?

DarNCelsius:
`theme=__aurorae__svg__Nord-Aurorae`

So not a bug then I guess?

LinusDierheimer:
Yes its your theme that is named weird

PR #70, opened 2021-06-26
Sanitize custom Plasma theme name

DarNCelsius:
Plasma prefixes the theme name in `kwinrc` with `"__aurorae__svg__"` for custom themes it seems.

The name shows up as normal in settings:
plasmatheme
Theme desktop entry:
```
[Desktop Entry]
Name=Nord-Aurorae
Comment=Nord-Aurorae
X-KDE-PluginInfo-Author=l4k1
X-KDE-PluginInfo-Email=l4k1987@gmail.com
X-KDE-PluginInfo-Version=1.0
X-KDE-PluginInfo-Name=Nord-Aurorae
X-KDE-PluginInfo-Category=
X-KDE-PluginInfo-Depends=
X-KDE-PluginInfo-License=GPLv3
X-KDE-PluginInfo-EnabledByDefault=true
```
The only place the string appears is in the `kwinrc` file, so I think #67 is valid after all.

### DIFF
--- a/src/modules/wmtheme.c
+++ b/src/modules/wmtheme.c
@@ -45,8 +45,16 @@ static void printWMThemeFromConfigFile(FFinstance* instance, const char* configF
         return;
     }
 
-    // Remove Plasma-generated prefix
-    ffStrbufRemoveStrings(&theme, 1, "__aurorae__svg__");
+    // Remove Plasma-generated prefixes
+    uint32_t idx = 0;
+
+    idx = ffStrbufFirstIndexS(&theme, "qml_");
+    if((idx != theme.length))
+        ffStrbufSubstrAfter(&theme, idx + 3);
+
+    idx = ffStrbufFirstIndexS(&theme, "svg__");
+    if((idx != theme.length))
+        ffStrbufSubstrAfter(&theme, idx + 4);
 
     printWMTheme(instance, theme.chars);
     ffStrbufDestroy(&theme);
--- a/src/modules/wmtheme.c
+++ b/src/modules/wmtheme.c
@@ -49,11 +49,11 @@ static void printWMThemeFromConfigFile(FFinstance* instance, const char* configF
     uint32_t idx = 0;
 
     idx = ffStrbufFirstIndexS(&theme, "qml_");
-    if((idx != theme.length))
+    if(idx != theme.length)
         ffStrbufSubstrAfter(&theme, idx + 3);
 
     idx = ffStrbufFirstIndexS(&theme, "svg__");
-    if((idx != theme.length))
+    if(idx != theme.length)
         ffStrbufSubstrAfter(&theme, idx + 4);
 
     printWMTheme(instance, theme.chars);
--- a/src/modules/wmtheme.c
+++ b/src/modules/wmtheme.c
@@ -45,6 +45,9 @@ static void printWMThemeFromConfigFile(FFinstance* instance, const char* configF
         return;
     }
 
+    // Remove Plasma generated garbage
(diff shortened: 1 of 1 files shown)


LinusDierheimer:
Looking at the neofetch source code, there actually is more that needs to be removed:
```awk
gsub(/theme=.*qml_|theme=.*svg__/,"",$0);
```
Can you implement that too?

DarNCelsius:
I'll have a look.

The second pattern is matching the same as this I guess. The Plastik theme is old and has `"kwin4_decoration_qml_"` prefixed, so would probably fit the first.

This is Neofetch with Plastik:

LinusDierheimer:
".*" is a regex expression to match a string of any length

DarNCelsius:
Having a hard time figuring out the ffsubstring routines.

With this string: `kwin4_decoration_qml_plastik` in "theme", calling
```c
idx = ffStrbufFirstIndexAfterS(&theme, 0, "qml_");
```
and then
```c
ffStrbufSubstrAfter(&theme, idx);
```
Should produce what?

LinusDierheimer:
`qml_plastik` i guess. Also you can use `ffStrbufFirstIndexS` instead of using start index 0

DarNCelsius:
It returns `"in4_decoration_qml_plastik"`

LinusDierheimer:
Will look into this.

DarNCelsius:
Shouldn't it produce "plastik" though? Since it's the index _after_?

LinusDierheimer:
No, the after is means it starts searching after a given index. Currently doing a rewrite of those functions with better naming.

DarNCelsius:
I see. Is there currently a substring function that would produce the string after (or before) the string you search for?

LinusDierheimer:
There currently isn't, the combination of the two methods is the best method currently. If you feel like it is needed, you can write a `ffStrbufRemoveStartS(const FFstrbuf* strbuf, const char* str)` or similar. 
I pushed a commit that renames some strbuf methods, the method u asked about is now called `ffStrbufNextIndexS`. Hope this makes things clearer.

DarNCelsius:
The combination of the two isn't producing that result though...

LinusDierheimer:
This works:
```c
FFstrbuf theme;
ffStrbufInit(&theme);
ffStrbufAppendS(&theme, "kwin4_decoration_qml_plastik");

uint32_t index = ffStrbufFirstIndexS(&theme, "qml_");
ffStrbufSubstrAfter(&theme, index + 3); //index of q plus ml_
ffStrbufPutTo(&theme, stdout);
```
Prints `plastik`.

DarNCelsius:
So yes then. :+1:

DarNCelsius:
New version works with both and no matches

src/modules/wmtheme.c
```diff
@@ -45,6 +45,17 @@ static void printWMThemeFromConfigFile(FFinstance* instance, const char* configF
         return;
     }
 
+    // Remove Plasma-generated prefixes
+    uint32_t idx = 0;
+
+    idx = ffStrbufFirstIndexS(&theme, "qml_");
+    if((idx != theme.length))
```
LinusDierheimer, line 52: Why the double brackets?

DarNCelsius: Oops. Copy paste error. I was going for a one liner prior. Something like
```c
if((idx = ffStrbufFirstIndexS(&theme, "qml_") != theme.length))
```
Couldn't get it to work.

LinusDierheimer: Probably because the index methods remove the index of the terminating null byte if the search failed.

LinusDierheimer:
Thank you.

PR #71, opened 2021-06-26
LXQT cursor support

DarNCelsius:


src/modules/cursor.c
```diff
@@ -105,6 +105,31 @@ static void printCursorXFCE(FFinstance* instance)
...
+    ffStrbufInit(&cursorTheme);
+
+    FFstrbuf cursorSize;
+    ffStrbufInit(&cursorSize);
+
+    if(ffParsePropFileConfigValues(instance, "/lxqt/session.conf", 2, (FFpropquery[]) {
+        {"cursor_theme =", &cursorTheme},
+        {"cursor_size =", &cursorSize}
+    })) {
+    }
```
LinusDierheimer, line 120: I see you copied this over from plasma. `ffParsePropFileConfigValues` return true if the file was found. So i use the if to set the default values when the file exists. This needs to be done in plasma, because default values are not saved. If this isn't the case in LXQt, a simple
```c
ffParsePropFileConfigValues(instance, "/lxqt/session.conf", 2, (FFpropquery[]) {
    {"cursor_theme =", &cursorTheme},
    {"cursor_size =", &cursorSize}
});
```
should be enough.

Also leading slash is not needed in path.

src/modules/cursor.c
```diff
@@ -105,6 +105,31 @@ static void printCursorXFCE(FFinstance* instance)
...
+
+    if(cursorTheme.length == 0)
+    {
+        ffPrintError(instance, FF_CURSOR_MODULE_NAME, 0, &instance->config.cursorKey, &instance->config.cursorFormat, FF_CURSOR_NUM_FORMAT_ARGS, "Couldn't find LXQt cursor in session.conf");
+        return;
```
LinusDierheimer, line 125: `cursorTheme` and `cursorSize` must be destroyed here. I admit i forgot that in the plasma code, can you add that there  too please?

src/modules/cursor.c
```diff
@@ -190,6 +215,12 @@ void ffPrintCursor(FFinstance* instance)
         return;
     }
 
+    if(ffStrbufStartsWithIgnCaseS(&wmde->dePrettyName, "LXQt"))
+    {
+        printCursorLXQt(instance);
+        return;
+    }
```
LinusDierheimer, line 222: Very similar to plasma code, maybe it is possible to write a single function which takes the path to the config file and the default values as arguments and use it for both?

DarNCelsius: I was just thinking that...

LinusDierheimer: If it should fall back to GTK you have two options: 
1) call `printCursorGTK` manually
2) return a bool value, for example like the env code

I would go for the second.

LinusDierheimer: If it doesn't has defaults and you want to write a single method, just pass `NULL` as default value. `ffStrbufAppendS` doesn't do anything if str is NULL. Maybe falling back to GTK makes sense for plasma too.

LinusDierheimer: Possible declaration of the function as a tip:
```c
bool printCursorFromConfigFile(FFinstance* instance, const char* relativeFilePath, const char* themeStart, const char* themeDefault, const char* sizeStart, const char* sizeDefault)
```

DarNCelsius: XCURSOR_THEME and XCURSOR_SIZE is not set on my system, so the env route seems unlikely.

LinusDierheimer: Yea this was just meant to be an example for how to handle functions which return bool

DarNCelsius: Ah. I should read better. :grin:

LinusDierheimer: Like this to be exactly:
```c
if(ffStrbufIgnCaseCompS(&wmde->dePrettyName, "KDE Plasma") == 0 && printCursorFromConfigFile(instance, "kcminputrc", "cursorTheme =", "Breeze", "cursorSize =", "24")
    return;
```

DarNCelsius: Defaults to [Adwaita](https://wiki.manjaro.org/index.php/LXQt#Change_Mouse_Cursor)

DarNCelsius:
Only tested on LXQt - could you check Plasma?

LinusDierheimer:
Works fine. However currently there is no GTK fallback. Do you think this is worth implementing?

DarNCelsius:
Adwaita is default for LXQt, and Breeze is default for Plasma, so I didn't see the need really.

If however, they fall back to gtk or something like env vars instead of using those defaults we should probably do something. I'm not sure how to check that.

I think we're good though.

LinusDierheimer:
Just thought about falling back if the file doesn't exists, but this is very unlikely if were running the corresponding session. You're right, we are good.

PR #80, opened 2021-07-17
Fixed Thunar terminal open segfault

### LINKED ISSUE #79: Thunar terminal open segfault
Fastfetch segfaults when opening a terminal from Thunar:
exobug
Command line:
`exo-open --working-directory %f --launch TerminalEmulator`

It works in PCManFM and PCManFM-Qt. I don't know the command line from those though

Subsequent fastfetch commands also segfaults at the same spot in that window.
Running `exo-open --launch TerminalEmulator` globally works fine.

DarNCelsius:
Fixes #79 - SEGFAULT on Openbox and  terminalExeName when terminal is launched from Thunar

### DIFF
--- a/src/common/detectTerminalShell.c
+++ b/src/common/detectTerminalShell.c
@@ -232,7 +232,6 @@ const FFTerminalShellResult* ffDetectTerminalShell(FFinstance* instance)
 
     ffStrbufInit(&result.terminalProcessName);
     ffStrbufInit(&result.terminalExe);
-    result.terminalExeName = result.terminalExe.chars;
 
     ffStrbufInit(&result.userShellExe);
     result.userShellExeName = result.userShellExe.chars;
@@ -241,6 +240,7 @@ const FFTerminalShellResult* ffDetectTerminalShell(FFinstance* instance)
     char ppid[256];
     snprintf(ppid, 255, "%i", getppid());
     getTerminalShell(&result, ppid);
+    result.terminalExeName = result.terminalExe.chars;
 
     getTerminalFromEnv(&result);
     getUserShellFromEnv(&result);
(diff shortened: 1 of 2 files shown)


DarNCelsius:
Is it your method or the change it complains about?

src/common/detectTerminalShell.c
```diff
@@ -232,7 +232,6 @@ const FFTerminalShellResult* ffDetectTerminalShell(FFinstance* instance)
 
     ffStrbufInit(&result.terminalProcessName);
     ffStrbufInit(&result.terminalExe);
-    result.terminalExeName = result.terminalExe.chars;
```
LinusDierheimer, line 235: I don't think this should be removed. I expect `result.terminalExeName` to always be a valid c-string. At this point it will _always_ point to the `'\0'` char of `result.terminalExe`. So setting it here can't really be the problem.

src/common/detectTerminalShell.c
```diff
@@ -241,6 +240,7 @@ const FFTerminalShellResult* ffDetectTerminalShell(FFinstance* instance)
     char ppid[256];
     snprintf(ppid, 255, "%i", getppid());
     getTerminalShell(&result, ppid);
+    result.terminalExeName = result.terminalExe.chars;
```
LinusDierheimer, line 243: This is definitely wrong, as it overwrites the `result.terminalExeName` set by `getTerminalShell`:
```c
ffStrbufAppendS(&result->terminalProcessName, name);
getProcessInformation(pid, &result->terminalProcessName, &result->terminalExe, &result->terminalExeName);
```
Which calls this code two methods later:
```c
uint32_t lastSlashIndex = ffStrbufLastIndexC(exe, '/');
if(lastSlashIndex < exe->length)
    *exeName = exe->chars + lastSlashIndex + 1;
```
Most probably your issue is something there.

LinusDierheimer (commented):
Sorry for the late response, i have been busy enrolling at university. It actually complains about the method being "too big", not about your change. Looking into disabling this warning later as in my opinion it us useless.

LinusDierheimer:
Closing this because of inactivity.

PR #91, opened 2021-10-16
Fix detecting Windows Terminal on WSL2

CarterLi:
Used logic from https://github.com/dylanaraps/neofetch/blob/master/neofetch#L3293

As for

```c
    if(result->terminalProcessName.length == 0)
        ffStrbufAppendS(&result->terminalProcessName, term);
```

Since we can only get here when `terminalProcessName` equals to `login` `systemd` `init` `0`, we should always use a better name

Result (with [patch applied](https://github.com/LinusDierheimer/fastfetch/issues/83#issuecomment-942446479)):
image


LinusDierheimer:
What happens when `SSH_CONNECTION` is set, but `SSH_TTY` is not, or empty?

CarterLi:
> What happens when `SSH_CONNECTION` is set, but `SSH_TTY` is not, or empty?

How can it happen?

Issue #83, opened 2021-07-29
Segmentation fault


AdisonCavani:
I have installed ``fastfetch`` from AUR on WSL2 (ArchLinux) and I'm getting "segmentation fault".
[Image](https://i.imgur.com/uXhuz3s.png)

LinusDierheimer:
Its hard to tell where the problem is just from the image, as fastfetch is multi threaded and any of those threads may segfault. Use `fastfetch --multithreading false` and tell me where it does stop now.

AdisonCavani:
error: XDG_RUNTIME_DIR not set in the environment.
[1]    4258 segmentation fault  fastfetch --multithreading false

LinusDierheimer:
This is more complicated as it looks, as i never use `XDG_RUNTIME_DIR`. The question where it stops remains.

LinusDierheimer:
This seems to be something specific with your setup, `XDG_RUNTIME_DIR= fastfetch` executes totally normal on my system, as it should. I don't have Windows installed so i can't test in WSL2. Without more information from you i can't do anything.

AdisonCavani:
![error](https://user-images.githubusercontent.com/76855180/131338250-cc0c3778-4b1c-435d-9774-8b7caf14ebc5.png)
This is how it looks on my WSL2

AdisonCavani:
After adding:
```
export XDG_RUNTIME_DIR=/some/directory/you/specify 
export RUNLEVEL=3
```
to .zshrc,
``XDG_RUNTIME_DIR not set in the environment`` is no longer visible, but I'm getting another error:

![error2](https://user-images.githubusercontent.com/76855180/131339156-c27cfff9-37fe-446a-9d22-66c1cbc18f11.png)

LinusDierheimer:
Resolution seems to be the problem. Fastfetch first tries to use on of the libraries in specified order:
* `libwayland-client.so`
* `libXrandr.so`
* `libX11.so`

Most probably you have one of the X11 libraries installed as a dependency, but no X Server running. This isn't a problem on linux, as i have implemented and tested this scenario, but maybe WSL2 behaves different. Can you look which of those libraries are present?

If resolution detection fails for all three, fastfetch falls back to scanning `/sys/class/drm/*/modes` itself. But this surely doesn't need `XDG_RUNTIME_DIR`, so the error is most probably in some X Configuration.

AdisonCavani:
[...]

PR #92, opened 2021-10-17
Add new module that shows local IP addresses which are useful with frontend development

CarterLi:
Main logic was used from https://stackoverflow.com/questions/212528/get-the-ip-address-of-the-machine

Screenshot with #91 and patch in #83 applied

image

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -116,6 +116,7 @@ set(SRCS
     src/modules/disk.c
     src/modules/battery.c
     src/modules/locale.c
+    src/modules/localip.c
     src/modules/colors.c
 )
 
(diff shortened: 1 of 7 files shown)


CarterLi:
> Pretty solid already, few things keeping me from merging.
> 
> What do you think about creating separate modules for IPV4, IPV6, etc? Would give better customization possibility, as a separate key / format string could be provided for each. For common code you could provide a helper function in src/common, similar to plasma for example.

To be honest I don't think anyone is interested in local IP of IPV6 ( but public IP ). Making a separate IPV6 module seems to be redundant

CarterLi:
Resolved

LinusDierheimer:
Remember you can permanently set a structure / other options in `~/.config/fastfetch/config.conf `

PR #96, opened 2021-10-19
Make it possible to build without optional dependencies

CarterLi:
For users that don't uses a desktop environment, for example WSL and Ubuntu Server, don't force them installing unnecessary dependencies that they will never use to build this awesome program.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,31 +43,55 @@ configure_file(src/fastfetch_config.h.in fastfetch_config.h)
 set(THREADS_PREFER_PTHREAD_FLAG ON)
 find_package(Threads REQUIRED)
 
-find_package (PkgConfig)
+find_package(PkgConfig REQUIRED)
 
-if(${PkgConfig_FOUND})
-    pkg_check_modules (GIO2 gio-2.0)
-    pkg_check_modules (DCONF dconf)
-endif(${PkgConfig_FOUND})
+pkg_check_modules (GIO2 gio-2.0)
+if(GIO2_FOUND)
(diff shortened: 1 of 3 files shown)


LinusDierheimer:
I thought about something like this:

```c
#if !defined(__has_include) || __has_include("wayland-client.h")
    #include "wayland-client.h"
    //Wayland code
#else
    #warning Building without Wayland support, since the header was not found!
    //Fallback code
#endif
```

LinusDierheimer:
Yea i just re read the documentation, seems like we have to use nested ifs: https://gcc.gnu.org/onlinedocs/gcc-10.1.0/cpp/_005f_005fhas_005finclude.html

LinusDierheimer:
I didn't include the xfconf header, since i only need one line from it, which i just copied:
```c
typedef struct _XfconfChannel XfconfChannel; // /usr/include/xfce4/xfconf-0/xfconf/xfconf-channel.h#L39
```
I did that, because i thought requiring everybody to pull in the xfce ecosystem, for a feature only needed in xfce environments, is a bit unnecessary, considering i only need this one line. Now that we will have support to build without a feature, i think it is better to remove the line and include the header.

LinusDierheimer:
Will you do the messages, or should i?

CarterLi:
Screenshot_2021-10-20-01-13-34-375_com termux

Program stucked at ffPrintPackages. Details unknown

CarterLi:
Suggestion: Can we arm a SIGINT handler to restore the cursor? Everytime I terminate the program with ^C the cursor disappear

PR #102, opened 2021-11-01
fix: check if `processName` is NULL/empty to prevent segfaults

NBonaparte:
fastfetch segfaults when using Wayland with river and sway on my system. I found that `processName` (i.e. `sessionDesktop`) in `applyPrettyNameIfWM` is NULL (as in https://github.com/LinusDierheimer/fastfetch/issues/83#issuecomment-942442166), so this adds a check to prevent the segfault like `applyPrettyNameIfDE` does.


LinusDierheimer:
Thanks. I don't know how it can be `NULL`  since every process in Linux is guaranteed to have a name, but the check doesn't hurt anything so i will merge it. Will investigate more so maybe ill find the root cause of it.

PR #103, opened 2021-11-03
Make the preset neofetch more like neofetch

Linus789:
- os-format: Print architecture without square brackets
- resolution-format: Do not print the refresh rate
- de-format: Use pretty name and print version
- theme-format: Print color scheme without brackets
- memory-format: Use MiB instead of Mib


LinusDierheimer:
Thanks. I don't sync changes i make in fastfetch with all presets, so this happens over time.

PR #123, opened 2022-01-22
Add packaging for Debian based distributions.

x-zvf:
This commit adds a script (`create-deb-pkg.sh`), which creates a deb package containing the fastfetch and flashfetch executables as well as the bash-completion file. The script must be executed in an environment, where `dpkg-deb` is present (a debian based environment).
Because Deb versioning is not compatible with fastfetch version strings, it is adapted to the format: `0:1$FF_VERSION-0`.

### DIFF
--- a/packaging/deb/create-deb-pkg.sh
+++ b/packaging/deb/create-deb-pkg.sh
@@ -1,14 +1,14 @@
 #!/bin/sh
 
-cd "$(dirname "$0")"
+cd "$(dirname "$0")" || exit
 echo "cleanup"
 rm -rf ./build ./fastfetch fastfetch.deb
 
 echo "building fastfetch"
 mkdir build
-cd build
+cd build || exit
 cmake ../../..
-cmake --build . -j$(nproc)
+cmake --build . --target fastfetch "-j$(nproc)"
 
 echo "generating package contents"
 # deb versions must start with a number, hence the 1 added before the rXXX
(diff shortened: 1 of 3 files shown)


LinusDierheimer:
Presets should be installed to `/usr/share/fastfetch/presets/`, compare with [PKGBUILD](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=fastfetch-git) for Archlinux.

Packaging flashfetch doesn't make any sense, as it is meant to be configured only at compile time.

packaging/deb/create-deb-pkg.sh
```diff
@@ -0,0 +1,32 @@
...
+
+cd "$(dirname "$0")"
+echo "cleanup"
+rm -rf ./build ./fastfetch fastfetch.deb
+
+echo "building fastfetch"
+mkdir build
+cd build
+cmake ../../..
+cmake --build . -j$(nproc)
```
LinusDierheimer, line 11: `cmake --build . --target fastfetch`. -j should be configured by the user with the MAKEFLAGS environment variable, not forced by default.

LinusDierheimer:
and adjust .gitignore, so no one can commit build files / binary blobs by default.

packaging/deb/create-deb-pkg.sh
x-zvf: Makes sense. Done

x-zvf:
Applied the suggested changes

LinusDierheimer:
LGTM, thank you.

LinusDierheimer:
Just noticed Installed-Size is hardcoded, is there a way to set it while building?

PR #130, opened 2022-02-08
Make dependencies check optional

ceamac:
Hi,

I need to be able to disable checking for dependencies.  This patch keeps the current default of checking them automatically.


CMakeLists.txt
```diff
@@ -48,82 +48,116 @@ find_package(Threads REQUIRED)
 
 find_package(PkgConfig REQUIRED)
```
LinusDierheimer, line 49: How does checking for PkgConfig work on your system, but not checking for a PkgConfig package?

LinusDierheimer (commented):
I'm not against merging this, i just want to know the why, as i have a feeling there is a better solution to the original problem than this.

CMakeLists.txt
ceamac: PkgConfig is required, I didn't change this line.  I only changed the optional dependencies.

LinusDierheimer: Yes that is what i was wondering about. If PkgConfig is available, what is the problem of including the headers in the binary build?

ceamac:
About the why.  I've written an ebuild for Gentoo.  Automagical dependencies are quite frowned upon in the Gentoo community, mostly because with the same ebuild and the same USE flags users may obtain different results if they happen to have some apparently unrelated dependencies installed or not.  Adding all dependencies unconditionally is not ok, if I make fastfetch depend on X or GNOME it cannot be used on minimal (maybe headless) installations.  So I need a way to disable a test if the user does not enable a certain USE flag, even if the corresponding dependency is actually present on that system.

You can see the old ebuild and the new ebuild here:

https://gitweb.gentoo.org/repo/proj/guru.git/tree/app-misc/fastfetch/fastfetch-0.569.ebuild?id=d85485e0f12f38fa8f06f15e3859b56f9b03f4c7
https://gitweb.gentoo.org/repo/proj/guru.git/tree/app-misc/fastfetch/fastfetch-0.569-r1.ebuild?h=dev&id=61a59bd4bd8f1bc7dd41dc0bff153c866d9ddb53

The second one looks nicer :)

ceamac:
Also because Gentoo does not support version numbers that start with r, I set the version to 0.569 for r569.

CMakeLists.txt
ceamac: I think if you include the needed headers with fastfetch it would actually work, because you dlopen the libs anyway.  The only problem was requiring big dependencies to be installed for a header that is only used at build time.

ceamac:
I see you merged it, thank you

LinusDierheimer:
So maybe this is where gentoo and arch defers. In arch those dependencies are listed as "make dependencies", so they are automatically installed during build time and automatically uninstalled afterwards. That is why fastfetch checks at runtime, if a library is available.

PR #131, opened 2022-02-10
Add packages support for Gentoo Linux

ceamac:
Support for counting installed packages on Gentoo

Caveat: runs an external command, so maybe it would be better to add an ifdef (or not add the support at all).

### DIFF
--- a/presets/verbose
+++ b/presets/verbose
@@ -5,7 +5,7 @@
 --kernel-format "Sysname: {}; Release: {}; Version: {}"
 --uptime-format "Days: {}; Hours: {}; Minutes: {}; Seconds: {}"
 --processes-format "Count: {}"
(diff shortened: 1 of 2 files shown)


LinusDierheimer:
TBH i don't like to open a new process for this. Is it possible to implement emerge package count by using one of the counting functions in package.c?

LinusDierheimer:
Much appreciated, thank you for your work!

LinusDierheimer:
I rewrote some of your logic using the strbuf methods fastfetch has built in, can you please confirm that it still works?

ceamac:
Yes, it still works.  Same result as before

LinusDierheimer:
Great. Thank you again for figuring the logic out.

SigHunter:
thank you very much for this, finally fastfetch is featurewise on par and I can switch from neofetch

PR #140, opened 2022-03-08
Adding RebornOS to the supported logos

shivanandvp:
The current PR aims to add support for RebornOS' logo to fastfetch
Screenshot_20220308_110801


shivanandvp:
Please feel free to replace it with a better logo if you can. I am not sure this is the best it can be. It could use improvement

LinusDierheimer:
Is it intended, that the keys don't line up in a vertical line?
Usually i copy the logos from neofetch, so i'm sure i can't create a better one.

shivanandvp:
@LinusDierheimer I could have it either line up or not. What do you prefer?

LinusDierheimer:
line up, so it is consistent with all other logos (and fetch programs)

LinusDierheimer:
LGTM

shivanandvp:
Screenshot_20220308_113735

shivanandvp:
Thanks 👍🏽

PR #168, opened 2022-06-08
Fix build if ImageMagick is not installed

Yureien:
The current build fails if ImageMagick is not installed, since the return type wasn't changed to `bool`. This PR fixes it.

BTW, I'm packaging fastfetch for NixOS: https://github.com/NixOS/nixpkgs/pull/176819/

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.1.0) # Threads::Threads
 
 project(fastfetch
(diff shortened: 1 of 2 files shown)


LinusDierheimer:
Oh thanks, that slipped through. Can you please bump the minor version number in CMakeLists.txt too? We shouldn't ship a version with this bug as latest.

And thanks for packaging on NixOS!

Yureien:
Done, bumped it.

PR #169, opened 2022-06-10
fix: correct "hout" to "hour"

ADawesomeguy:
I just made this PR because I saw it and it was convenient.


LinusDierheimer:
Thanks

PR #176, opened 2022-06-15
Quality of life improvement: make run.sh not run ff, if the build fails

x-zvf:
If `set -e` is used, the script will not run `build/fastfetch` if the build failed.


LinusDierheimer:
Looks fine.

PR #180, opened 2022-06-17
Add Alpine Linux logos

draumaz:
Logos from @dylanaraps' neofetch and pfetch projects. Tested working on latest Alpine x86-64 build.


LinusDierheimer:
Thanks a lot.

PR #189, opened 2022-07-07
Fix memory leak in ffParsePropFileValues

x-zvf:
If `ffParsePropFileValues` gets called with `numQueries > 4`, we allocate `unsetValues` the heap, which does not get freed, if the file can not be opened.


LinusDierheimer:
Can we just do the fopen before the allocation? I think code would be cleaner that way. Thanks for finding this leak!

x-zvf:
The commit `7bc85b2` refactors the function to use cleanup with `goto`, but you are going to have to decide if you actually find that cleaner, or I should revert. Personally I prefer the goto style, but it is not consistent with other parts of the code.

LinusDierheimer:
LGTM

PR #191, opened 2022-07-12
[enhancement] Add best effort terminal font detection for alacritty.

x-zvf:
Alacritty's config files are technically YML files, so
simply looking for a prop value "family:" may break some of the time,
but is in fact better than what neofetch does and much easier
to implement then writing a YML parser :)

### DIFF
--- a/src/modules/terminalfont.c
+++ b/src/modules/terminalfont.c
@@ -53,7 +53,6 @@ static void printTerminalFontFromConfigFile(FFinstance* instance, const char* co
     FFstrbuf fontName;
     ffStrbufInit(&fontName);
     ffParsePropFileConfig(instance, configFile, start, &fontName);
-
     if(fontName.length == 0)
         ffPrintError(instance, FF_TERMFONT_MODULE_NAME, 0, &instance->config.terminalFont, "Couldn't find terminal font in \"$XDG_CONFIG_HOME/%s\"", configFile);
     else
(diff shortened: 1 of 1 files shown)


src/modules/terminalfont.c
```diff
@@ -179,6 +178,26 @@ static void printXCFETerminal(FFinstance* instance)
...
+    ffParsePropFileConfig(instance, "alacritty/alacritty.yml", "family:", &fontName);
+    if(fontName.length == 0)
+        ffParsePropFileConfig(instance, "alacritty.yml", "family:", &fontName);
+    if(fontName.length == 0)
+        ffParsePropFileConfig(instance, ".alacritty.yml", "family:", &fontName);
+    if(fontName.length == 0)
+        ffStrbufAppendS(&fontName, "alacritty"); //by default alacritty uses it's own font called alacritty
+
+    FFfont font;
+    ffFontInitPango(&font, fontName.chars);
```
LinusDierheimer, line 194: Does alacritty really uses [pango](https://pango.gnome.org/) for font IO? This is usually a GTK thing. Pango fonts for example look like this: `Semi-Bold NotoMono Underline 10`. According to the one and only [arch wiki](https://wiki.archlinux.org/title/Alacritty) alacritty just saves the name under family, so a `ffFontInitCopy` should be much faster and enough.
The key `size: ` should only be used for font size, so we can parse for it too. You can use `ffParsePropFileConfigValues` to parse for multiple values at once.

x-zvf: No it does not, I just copy-and-pasted :).
I'll rewrite to use ffFontInitCopy

src/modules/terminalfont.c
```diff
@@ -179,6 +179,40 @@ static void printXCFETerminal(FFinstance* instance)
...
+    if(fontName.length == 0 || fontSize.length == 0)
+        ffParsePropFileConfigValues(instance, "alacritty.yml", 2, fontQuery);
+    if(fontName.length == 0 || fontSize.length == 0)
+        ffParsePropFileConfigValues(instance, ".alacritty.yml", 2, fontQuery);
+    if(fontName.length == 0 || fontSize.length == 0)
+    {
+         //by default alacritty uses it's own font called alacritty at size 11
+        ffStrbufAppendS(&fontName, "alacritty");
+        ffStrbufAppendS(&fontSize, "11");
+    }
```
LinusDierheimer, line 204: We should check for font and size independently, for example if font found, and size wasn't, this would override the font.

src/modules/terminalfont.c
```diff
@@ -179,6 +179,40 @@ static void printXCFETerminal(FFinstance* instance)
...
+
+    FFfont font;
+    ffFontInitCopy(&font, fontName.chars);
+    ffStrbufInitCopy(&font.size, &fontSize);
```
LinusDierheimer, line 208: `font.size` is already initialized by fontInitCopy. A simple `ffStrbufAppend` should do the trick. Otherwise we leak memory.

src/modules/terminalfont.c
x-zvf: good point

PR #205, opened 2022-07-27
Fix build error on Android

licy183:
host.c has an #elif, but actually #else is needed.
Related issue: https://github.com/termux/termux-packages/issues/11361.


LinusDierheimer:
Thank. Do you need a version bump for this? If yes, please increase project version in CMakeLists.txt. Github will automatically create a new release and tag.

PR #226, opened 2022-09-10
OpenGL: add macOS support with SDL2

CarterLi:
SDL is more portable and may work with future windows version.


LinusDierheimer:
TBH i'd rather not merge this. I think OpenGL Context creation on MacOS should be done using CGL. It is not only guaranteed to be available (we don't need the whole FF_HAVE_* stuff), but also very simple to set up, according to [this](https://stackoverflow.com/questions/11383510/setting-up-an-opengl-context-with-cgl-on-mac-os-x) SO post.

About the future windows version, i doubt that the future is in the foreseeable time. Fastfetch was designed heavily around the posix API, to the point that there isn't a single file that would already compile on windows. If support for it is needed, we would need to create a whole abstraction layer around things like IO, threading, dynamic library loading, etc.

CarterLi:
[Windows does support lots of posix APIs](https://docs.microsoft.com/en-us/cpp/c-runtime-library/low-level-i-o?view=msvc-170). With `_CRT_NONSTDC_NO_WARNINGS` pre-defined, `open` `write` `stat` can be used without modifications.

Windows does support ANSI escape code.

Windows (say MinGW-w64) does have a pthread implementation ( performs not very well though ). A better option is just disable multithreading on Windows like what you did on Android and macOS.

There are dlopen like API libraries on Windows, and it's trivial to write a `LoadLibrary` based implementations.

Support the common things on Windows is not very hard IMO.

CarterLi:
> TBH i'd rather not merge this. I think OpenGL Context creation on MacOS should be done using CGL. It is not only guaranteed to be available (we don't need the whole FF_HAVE_* stuff), but also very simple to set up, according to [this](https://stackoverflow.com/questions/11383510/setting-up-an-opengl-context-with-cgl-on-mac-os-x) SO post.

The entire SDL2 part is fully optional. We had EGL, GLX, OSMESA, have another option is not that bad.

CarterLi:
I'll remove the SDL part if you insist. This PR also generally simplifies the usage age of library-loading macros. Let me know what you think.

src/modules/opengl.c
```diff
@@ -372,6 +452,11 @@ static const char* glPrint(FFinstance* instance)
             error = glxPrint(instance);
     #endif
 
+    #ifdef FF_HAVE_SDL2
+        if(error != NULL)
+            error = sdl2Print(instance);
+    #endif
+
```
LinusDierheimer, line 459: SDL will always use CGL on MacOS, so i don't see a need to go through its abstraction.

CMakeLists.txt
```diff
@@ -69,6 +70,9 @@ endif()
 
 set(CMAKE_C_STANDARD 11)
 set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wextra -Wconversion")
+if(APPLE AND DEFINED ENV{HOMEBREW_PREFIX})
+    set(CMAKE_EXE_LINKER_FLAGS "-Wl,-rpath,$ENV{HOMEBREW_PREFIX}/lib")
+endif()
```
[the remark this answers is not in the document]
CarterLi: It's used for dlopen finding dylibs installed by homebrew. `/opt/homebrew/lib` is not on dlopen search path by default

CarterLi:
SDL code removed

.github/workflows/push.yml
```diff
@@ -79,7 +79,7 @@ jobs:
         uses: github/codeql-action/analyze@v2
 
       - name: run fastfetch
-        run: ./fastfetch --recache --disable-linewrap false --hide-cursor false --show-errors true
+        run: ./fastfetch --recache --disable-linewrap false --hide-cursor false --show-errors true --load-config ../presets/all
```
LinusDierheimer, line 82: GH actions builds in the source code directory, so the path is just `presets/all`

CarterLi: fixed

LinusDierheimer:
Thanks.

PR #230, opened 2022-09-13
OpenGL / OpenCL: add macOS support

CarterLi:
Note: FF_LIBRARY_LOAD and friends are not used because loading framework bundles are not supported by FF_LIBRARY_LOAD and support it is not easy. Since CGL is guaranteed to be supported by macOS system, we simply use static linking in this case.

image

### DIFF
--- a/src/modules/opengl.c
+++ b/src/modules/opengl.c
@@ -7,7 +7,7 @@
 #define FF_OPENGL_MODULE_NAME "OpenGL"
 #define FF_OPENGL_NUM_FORMAT_ARGS 4
 
-#if defined(FF_HAVE_EGL) || defined(FF_HAVE_GLX) || defined(FF_HAVE_OSMESA)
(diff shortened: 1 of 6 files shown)


src/modules/opengl.c
```diff
@@ -337,6 +337,45 @@ static const char* osMesaPrint(FFinstance* instance)
...
+        return "CGLChoosePixelFormat() failed";
+    }
+
+    CGLContextObj ctx;
+    if (CGLCreateContext(pix, NULL, &ctx) != kCGLNoError) {
+        return "CGLCreateContext() failed";
+    }
+    if (CGLSetCurrentContext(ctx) != kCGLNoError) {
+        return "CGLSetCurrentContext() failed";
+    }
```
LinusDierheimer, line 363: This leaks the pixel format

LinusDierheimer: I*ll fix by myself.

PR #233, opened 2022-09-15
Fix bash completions file

ceamac:
This is a trivial fix, the bash completions file in 1.7.1 cannot be sourced

If you don't want to make a new release for a 1 line change I can revert CMakeLists and CHANGELOG.


LinusDierheimer:
Thanks, reminds me not to test only with ZSH...

PR #236, opened 2022-09-16
Various fixes

CarterLi:
image

### DIFF
--- a/src/detection/battery/battery_apple.c
+++ b/src/detection/battery/battery_apple.c
@@ -13,7 +13,7 @@ const char* ffDetectBatteryImpl(FFinstance* instance, FFlist* results)
         return "IOServiceMatching(\"AppleSmartBattery\") failed";
 
     io_iterator_t iterator;
-    if(IOServiceGetMatchingServices(0, matchDict, &iterator) != kIOReturnSuccess)
+    if(IOServiceGetMatchingServices(MACH_PORT_NULL, matchDict, &iterator) != kIOReturnSuccess)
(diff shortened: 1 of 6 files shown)


LinusDierheimer:
Can we strip `Apple` and `Intel` prefixes from the GPU name and set vendor accordingly? Would not only give a prettier output, but format strings would work better too.

Personally i would also remove the `Graphics` suffix.

CarterLi:
Problem is that the string format returned is not guaranteed. What about AMD graphic cards?

LinusDierheimer:
I would write code similar to this (don't blindly copy it, i am writing this directly on github):

```c
ffStrbufRemoveIgnCaseEndS(&name, "Graphics");
ffStrbufTrimRight(&name);

char intelPrefix[] = "AppleIntel";
char applePrefix[] = "Apple";

if(ffStrbufStartsWithIgnCaseS(&name, intelPrefix)
{
    ffStrbufSubstrAfter(&name, sizeof(intelPrefix);
    ffStrbufAppendS(&vendor, "Intel");
}
else if(ffStrbufStartsWithIgnCaseS(&name, applePrefix)
{
    ffStrbufSubstrAfter(&name, sizeof(applePrefix);
    ffStrbufAppendS(&vendor, "Apple");
}

ffStrbufTrimLeft(&name);
```

CarterLi:
Updated, now we can correctly detect Intel Iris graphics

image

As for vendor, I don't think I want to do it before I can verify whether it works for other graphic cards.

LinusDierheimer:
Thank you.

PR #238, opened 2022-09-17
TerminalFont: move detection code into `src/detection`

CarterLi:
In order to return a constant string ( to avoid memory allocation ), all string formatting code ( mostly in `terminalfont_linux.c` ) was removed.

The resource owner of `FFfont* font` was not very clear. For now `font` must be inited inside `detection` code, but it must be destroyed outside.

Any suggestions are welcome, or you may merge my PR first then change them yourself


CarterLi:
To be honest resource management in C is really hard. I still prefer C++ for RAII than C

LinusDierheimer:
Sorry, i saw your pull request to late, and did it myself. Thanks anyway.

PR #242, opened 2022-09-19
Media: support song detection on macOS

CarterLi:
Will do player tomorrow if I have time

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -189,7 +189,6 @@ set(LIBFASTFETCH_SRC
     src/detection/gtk.c
     src/detection/terminalShell.c
     src/detection/vulkan.c
-    src/detection/media.c
     src/detection/datetime.c
     src/detection/temps.c
     src/detection/title.c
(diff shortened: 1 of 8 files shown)


LinusDierheimer:
I can't test it, but code wise it seems fine. Thanks a lot!

LinusDierheimer:
I noticed that you used objective c, but i don't see anything that requires the use of it. Why not stay in c?

And is the media remote framework guaranteed to exist on MacOS? If yes we can just link against it.

CarterLi:
> I noticed that you used objective c, but i don't see anything that requires the use of it. Why not stay in c?

using of `<CoreFoundation/CoreFoundation.h>` requires obj-c ( which is a obj-c header file )

Also notice the usage of `^` block syntax ( may also be provided as a C extension )

> 
> And is the media remote framework guaranteed to exist on MacOS?

Yes.

> If yes we can just link against it.

No. `/System/Library/PrivateFrameworks/MediaRemote.framework/MediaRemote` is a private framework, which cannot be directly linked against. `MediaRemote.framework` isn't in the framework search path, `-framework /System/Library/PrivateFrameworks/MediaRemote.framework` doesn't work and `MediaRemote.framework` provides no header files.

LinusDierheimer:
Then i am completely fine with the current state.

PR #245, opened 2022-09-20
[macOS] add player detection

CarterLi:
image


CarterLi:
Notice the error `Failed to detect memory`

That was because `sysctl vm.swapusage` return 0

> vm.swapusage: total = 0.00M  used = 0.00M  free = 0.00M  (encrypted)

which is an expected behavior because swap is dynamic allocated in macOS. In this case, using -1 as the error identifier should be a better option

LinusDierheimer:
Thanks for your work on this.

PR #246, opened 2022-09-20
WMTheme: support macOS

CarterLi:
Note the `AppleAccentColor` part was not tested because the `AppleAccentColor` doesn't exist in the plist file

### DIFF
--- a/src/modules/wmtheme.c
+++ b/src/modules/wmtheme.c
@@ -229,6 +229,87 @@ static void printOpenbox(FFinstance* instance, const FFstrbuf* dePrettyName)
     ffStrbufDestroy(&absolutePath);
 }
 
+#ifdef FF_HAVE_LIBPLIST
+#include "common/library.h"
+#include "common/io.h"
+#include <plist/plist.h>
+
+static const char* quartzCompositorParsePlist(FFinstance* instance, const FFstrbuf* content, FFstrbuf* theme) {
+    FF_LIBRARY_LOAD(libplist, &instance->config.libplist, "dlopen libplist failed", "libplist-2.0"FF_LIBRARY_EXTENSION, 2);
+    FF_LIBRARY_LOAD_SYMBOL_MESSAGE(libplist, plist_is_binary);
+    FF_LIBRARY_LOAD_SYMBOL_MESSAGE(libplist, plist_from_bin);
+    FF_LIBRARY_LOAD_SYMBOL_MESSAGE(libplist, plist_from_xml);
+    FF_LIBRARY_LOAD_SYMBOL_MESSAGE(libplist, plist_dict_get_item);
+    FF_LIBRARY_LOAD_SYMBOL_MESSAGE(libplist, plist_get_string_ptr);
+    FF_LIBRARY_LOAD_SYMBOL_MESSAGE(libplist, plist_get_uint_val);
+    FF_LIBRARY_LOAD_SYMBOL_MESSAGE(libplist, plist_free);
+
+    plist_t root_node = NULL;
+    if (ffplist_is_binary(content->chars, content->length))
+        ffplist_from_bin(content->chars, content->length, &root_node);
+    else
+        ffplist_from_xml(content->chars, content->length, &root_node);
+
+    if(root_node == NULL)
+        return "parsing Quartz Compositor preference file failed";
+
+    plist_t pWmThemeColor = ffplist_dict_get_item(root_node, "AppleAccentColor");
+    uint64_t wmThemeColor = 100;
+    if (pWmThemeColor != NULL)
+        ffplist_get_uint_val(pWmThemeColor, &wmThemeColor);
+    switch (wmThemeColor)
+    {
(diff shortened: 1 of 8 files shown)


src/modules/wmtheme.c
```diff
@@ -229,6 +229,87 @@ static void printOpenbox(FFinstance* instance, const FFstrbuf* dePrettyName)
...
+        ffplist_get_uint_val(pWmThemeColor, &wmThemeColor);
+    switch (wmThemeColor)
+    {
+        case (uint64_t)-1: ffStrbufAppendS(theme, " (Graphite)");
+        case 0: ffStrbufAppendS(theme, "Red");
+        case 1: ffStrbufAppendS(theme, "Orange");
+        case 2: ffStrbufAppendS(theme, "Yellow");
+        case 3: ffStrbufAppendS(theme, "Green");
+        case 5: ffStrbufAppendS(theme, "Purple");
+        case 6: ffStrbufAppendS(theme, "Pink");
```
LinusDierheimer, line 268: This needs break statements.

src/util/FFstrbuf.c
```diff
@@ -29,6 +29,14 @@ void ffStrbufInitCopy(FFstrbuf* strbuf, const FFstrbuf* src)
     ffStrbufAppend(strbuf, src);
 }
 
+void ffStrbufInitS(FFstrbuf* strbuf, const char* str)
+{
+    uint32_t bufSize = (uint32_t)strlen(str) + 1;
+    ffStrbufInitA(strbuf, FASTFETCH_STRBUF_DEFAULT_ALLOC > bufSize ? FASTFETCH_STRBUF_DEFAULT_ALLOC : bufSize);
+    memcpy(strbuf->chars, str, bufSize);
+    strbuf->length = bufSize - 1;
```
LinusDierheimer, line 37: Is this really faster than an init and an append?

CarterLi: Ok, but isn't this a bug?

https://github.com/LinusDierheimer/fastfetch/blob/master/src/util/FFstrbuf.c#L100

CarterLi: https://github.com/LinusDierheimer/fastfetch/blob/master/src/util/FFstrbuf.c#L112

Shouldn't be `strbuf.length + length + 1`?

LinusDierheimer: No, `ffStrbufEnsureFrees` ensures that there is enough space to append the given length, including the null byte.

CarterLi:
Ok, I misunderstood.

However I'd say that the existing code is not optimised.

```c
void ffStrbufAppendS(FFstrbuf* strbuf, const char* value)
{
    if(value == NULL)
        return;

    for(uint32_t i = 0; value[i] != '\0'; i++)
    {
        if(i % 16 == 0)
            ffStrbufEnsureFree(strbuf, 16);
        strbuf->chars[strbuf->length++] = value[i];
    }

    strbuf->chars[strbuf->length] = '\0';
}
```

1. Reallocating in a loop is really not a good idea because memory allocating is slow. It's always better to pre-calculate the required buffer size.
2. `memcpy` is always faster than hand coded plain loop ( if compiler failed to optimise the loop to a memcpy call ). A well coded `memcpy` ( such as the one provided in glibc ) uses [SIMD instructions](https://en.wikipedia.org/wiki/Single_instruction,_multiple_data) copies 16 ( with SSE2 ) or 32 ( with AVX ) bytes per loop, and can be about 10~20 times faster than hand coded plain loop which copies only 1 byte per loop. The larger the buffer is, the faster memcpy can be.

Quick benchmark: https://quick-bench.com/q/rv4D5N9zynZRRcWb2Td7y0BVneI

image

LinusDierheimer:
Makes sense. Thanks for looking into this.

CarterLi:
ffStrbufAppendS can be further optimised by make it inline-able.

```c
static inline void ffStrbufAppendS(FFstrbuf* strbuf, const char* value)
{
    if(value == NULL)
        return;

    ffStrbufAppendNS(strbuf, (uint32_t)strlen(value), value);
}
```

If the value is a string literal, its length can be known at compile time. In which case `(uint32_t)strlen(value)` can be completely optimised away.

```c
// Before compiling
ffStrbufAppendS(&strbuf, "123456789");

// After compiling
ffStrbufAppendNS(&strbuf, 9, "123456789");
```

LinusDierheimer:
With the current implementation it might even make sense to force inline it.

CarterLi:
Quick test: https://gcc.godbolt.org/z/z3rrn9vfa

image

LinusDierheimer:
I don't know enough much about that, but i don't think this applies to fastfetch, because the method implementation is in another compilation unit. Might not be a problem with IPO tho. If we want to make sure it is getting inlined, we'd have to implement it as inline method in the header.

CarterLi:
Yes. A inline-able function must be in the header. Just move them into `FFstrbuf.h` you are done.

PR #247, opened 2022-09-21
[macOS] support cpuUsage and other improvements

CarterLi:
image

We now have almost full macOS support. I think it's time to make a homebrew package


LinusDierheimer:
LGTM, thanks a lot.

PR #251, opened 2022-09-23
PowerAdapter: make it a seperated module ( macOS )

CarterLi:
Also fixed bugs in FFstrbuf. BTW I think it's a good idea to make `ffStrbufInit` and `ffListInit` allocates no memory, which can save time when the buffer is not used ( conditionally )


LinusDierheimer:
Seems fine. About GH actions failing on MacOS, i'm working on that.

PR #253, opened 2022-09-23
[macOS] Disk & Media improvements

CarterLi:


LinusDierheimer:
Thanks, you are really pushing macos towards feature parity with linux!

LinusDierheimer:
BTW do you have any idea why github actions fails for macos at the moment? I can't reproduce it local, and reverting the chances didn't fix the segfault. Also when trying to build with debug mode on github, it does work fine...

CarterLi:
No I don't

PR #255, opened 2022-09-24
Fix GiBMiB format RAM units due to value being hardcoded in some presets

dr460nf1r3:


LinusDierheimer:
Ahh forgot them. Thanks!

LinusDierheimer (approved):

PR #257, opened 2022-09-24
Various improvements, see commits history for detail

CarterLi:


LinusDierheimer:
Seems good, i especially like the use of `assert()`, i really should start using it myself...
Thanks!

PR #258, opened 2022-09-25
[macOS] support Temps, Font; Host pretty print; other improvements

CarterLi:
image


CarterLi:
Compare to neofetch (hyfetch)

image

* Items that fastfetch doesn't support:

1. [GPU driver](https://github.com/hykilpikonna/hyfetch/blob/master/neofetch#L4605), which always prints `macOS Default Graphics Driver`

* Items that neofetch (hyfetch) doesn't support:

1. Font
2. CPU Usage ( cost 1 second though )
3. Swap
4. Date & Time
5. Vulkan
6. OpenGL
7. OpenCL

* Items that fastfetch does better

1. WM Theme: We correctly prints `Multicolor` as macOS suggested while neofetch prints `Blue`
2. Temperature: We now support macOS
3. Media: We queries macOS NowPlaying service, thus support most media players

src/detection/battery/battery_apple.c
```diff
@@ -1,9 +1,34 @@
...
+        return FF_BATTERY_TEMP_UNSET;
+
+    double result = 0;
+    for(uint32_t i = 0; i < temps.length; ++i)
+    {
+        FFTempValue* tempValue = (FFTempValue*)ffListGet(&temps, i);
+        result += tempValue->value;
+        //TODO: do we really need this?
+        ffStrbufDestroy(&tempValue->name);
+        ffStrbufDestroy(&tempValue->deviceClass);
```
LinusDierheimer, line 25: I don't think this belongs here, as this is also the place where the list is being destroyed.

src/modules/font.c
```diff
@@ -24,76 +16,35 @@ void ffPrintFont(FFinstance* instance)
...
-            {FF_FORMAT_ARG_TYPE_STRBUF, &gtk}
-        });
+        else
+        {
+            ffPrintFormat(instance, FF_FONT_MODULE_NAME, 0, &instance->config.font, FF_FONT_NUM_FORMAT_ARGS, (FFformatarg[]){
+                {FF_FORMAT_ARG_TYPE_STRING, font->type},
+                {FF_FORMAT_ARG_TYPE_STRBUF, &font->fontPretty}
+            });
+        }
+        ffStrbufDestroy(&font->fontPretty);
```
LinusDierheimer, line 46: I don't like the idea of printing QT and GTK in separate lines. Can we keep them inline?

src/modules/gpu.c
```diff
@@ -25,11 +26,14 @@ static void printGPUResult(FFinstance* instance, uint8_t index, FFcache* cache,
     if(gpu->coreCount != FF_GPU_CORE_COUNT_UNSET)
         ffStrbufAppendF(&output, " (%d)", gpu->coreCount);
 
+    if(!isnan(gpu->temperature)) //FF_GPU_TEMP_UNSET
```
LinusDierheimer, line 29: On my system, `isnan` requires linking `libm`, i don't think thats worth over simple `gpu->temperature != gpu->temperature`.

Same goes for the other `isnan` calls.

src/modules/gpu.c
```diff
@@ -25,11 +26,14 @@ static void printGPUResult(FFinstance* instance, uint8_t index, FFcache* cache,
...
+        ffStrbufAppendF(&output, " - %.1f°C", gpu->temperature);
+
     ffPrintAndAppendToCache(instance, FF_GPU_MODULE_NAME, index, &instance->config.gpu, cache, &output, FF_GPU_NUM_FORMAT_ARGS, (FFformatarg[]){
         {FF_FORMAT_ARG_TYPE_STRBUF, &gpu->vendor},
         {FF_FORMAT_ARG_TYPE_STRBUF, &gpu->name},
         {FF_FORMAT_ARG_TYPE_STRBUF, &gpu->driver},
-        {FF_FORMAT_ARG_TYPE_DOUBLE, &gpu->temperature},
+        {FF_FORMAT_ARG_TYPE_DOUBLE, &gpu->temperature}, //FIXME: temperature shouldn't be cached
```
LinusDierheimer, line 36: Working on that right now, like i already did for CPU

LinusDierheimer (commented):
Amazing work!

LinusDierheimer:
We can also make `type` a part of the `FFfont` struct. Then the fond detection method can simply return a list of `FFfont`s.

LinusDierheimer:
Also i am of the opinion, that temperatures shouldn't be shown by default.

LinusDierheimer:
> As the latest version 1.7.2 still crashes on Intel MBP, what about a new release?

I want to finish the transition to the new caching mechanism, then i will do a new release.

src/detection/battery/battery_apple.c
CarterLi: I meant that although we got `name` and `deviceClass` but they were never displayed

LinusDierheimer: `name` and `deviceClass` are used on linux. 
`name` is used to identify which hwmon is the cpu, `deviceClass` is used by gpu, to map the device class found by libpci to the correct hwmon.

I don't think we need those on apple.

PR #259, opened 2022-09-25
New Ubuntu logo

AloneER0:
Maybe it's not really a useful pull request but i would like to know what do you think about the new Ubuntu ascii from neofetch master 
image

There shouldn't be any build errors.


LinusDierheimer:
I like the new logo, and it makes sense to use it, as it resembles the new ubuntu logo. But can we keep the old logo as `ubuntu_old` around? Similar to what for example fedora has.

AloneER0:
Oh yeah sure! i can bring back the old logo

AloneER0:
Let me push the changes.
image

LinusDierheimer:
Great, thank you!

AloneER0:
I just realized it has some formatting errors..

AloneER0:
let me correct them

PR #263, opened 2022-09-26
Support terminal font detection on Android and other fixes

CarterLi:
image

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -338,6 +338,11 @@ if(HAVE_SYSINFO_H)
     target_compile_definitions(libfastfetch PUBLIC FF_HAVE_SYSINFO_H)
 endif()
 
+CHECK_INCLUDE_FILE("utmpx.h" HAVE_UTMPX_H)
+if(HAVE_UTMPX_H)
+    target_compile_definitions(libfastfetch PRIVATE FF_HAVE_UTMPX_H)
+endif()
+
 function(ff_lib_enable VARNAME)
     if(NOT ENABLE_${VARNAME})
         return()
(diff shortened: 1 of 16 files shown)


src/modules/users.c
```diff
@@ -1,7 +1,15 @@
...
-#include <utmpx.h>
+#if __has_include(<utmpx.h>)
+    #include <utmpx.h>
+#else
+    //for Android compatibility
+    #include <utmp.h>
+    #define utmpx utmp
+    #define setutxent setutent
+    #define getutxent getutent
+#endif
```
LinusDierheimer, line 12: I'd prefer to do that with cmakes `CHECK_INCLUDE_FILE`, too keep the non standard compliance low.

CMakeLists.txt
```diff
@@ -338,6 +339,12 @@ if(HAVE_SYSINFO_H)
     target_compile_definitions(libfastfetch PUBLIC FF_HAVE_SYSINFO_H)
 endif()
 
+CHECK_INCLUDE_FILE("utmpx.h" HAVE_UTMPX_H)
+if(HAVE_UTMPX_H)
+    # needs to be public, because changes fastfech.h ABI
+    target_compile_definitions(libfastfetch PUBLIC FF_HAVE_UTMPX_H)
```
LinusDierheimer, line 345: This can be `PRIVATE` because it isn't used in a header included by `fastfetch.c` / `flashfetch.c`.

src/modules/cpuUsage.c
```diff
@@ -1,21 +1,55 @@
...
+    return (timeNow.tv_sec * 1000) + (timeNow.tv_usec / 1000);
+}
+
+static long inUseAll1, totalAll1, startTime;
+
+void ffPrepareCPUUsage()
+{
+    startTime = getTimeInMs();
+    ffGetCpuUsageInfo(&inUseAll1, &totalAll1);
+}
```
LinusDierheimer, line 22: This doesn't seem to be time expensive on my system, maybe we can just always start it somewhere in ffStart

LinusDierheimer:
Thanks

PR #271, opened 2022-09-30
Improvements

CarterLi:


CMakeLists.txt
```diff
@@ -97,6 +97,10 @@ endif()
 # Target FS structure #
 #######################
 
+if(CMAKE_INSTALL_PREFIX)
+    set(TARGET_DIR_ROOT "${CMAKE_INSTALL_PREFIX}")
+endif()
```
LinusDierheimer, line 102: We can't do that, for example on linux the install prefix often is `/usr` (see the [aur package](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=fastfetch-git)). 

If you install it without a package manager, the prefix usually is `/usr/local`, so we can't even use it for `TARGET_DIR_USR`.

CarterLi: ok I dropped this commit. But this is the official way that configures the prefix path, we should use it somehow

PR #274, opened 2022-10-01
macOS: respect CMAKE_INSTALL_SYSCONFDIR; use Cocoa API to parse plist files

CarterLi:


CMakeLists.txt
```diff
@@ -123,6 +122,10 @@ if(NOT TARGET_DIR_HOME)
     endif()
 endif()
 
+if(NOT TARGET_DIR_ETC)
+    set(TARGET_DIR_ETC "${CMAKE_INSTALL_SYSCONFDIR}")
+endif()
```
LinusDierheimer, line 127: This does not work if `CMAKE_INSTALL_SYSCONFDIR` is not given during configuring. It is set by `include(GNUInstallDirs)`, which must be called after we set our default values.

`TARGET_DIR_ETC` and `CMAKE_INSTALL_SYSCONFDIR` will always be the same, so i would do the following:

Line 125 - 127:
```cmake
if(NOT CMAKE_INSTALL_SYSCONFDIR)
    set(CMAKE_INSTALL_SYSCONFDIR "${TARGET_DIR_ROOT}/etc" CACHE PATH "..." FORCE)
endif()
```
fastfetch_config.h.in:
```c
#define FASTFETCH_TARGET_DIR_ETC "@CMAKE_INSTALL_SYSCONFDIR@"
```

LinusDierheimer (commented):
I'll trust you on the apple code.

CMakeLists.txt
CarterLi: > TARGET_DIR_ETC and CMAKE_INSTALL_SYSCONFDIR will always be the same

It's not the case if a user specifies `CMAKE_INSTALL_SYSCONFDIR`. It's necessary to make it work with macOS homebrew

CarterLi: Source: https://github.com/Homebrew/homebrew-core/blob/master/Formula/fastfetch.rb

CarterLi:
> I'll trust you on the apple code.

Thanks.

CMakeLists.txt
LinusDierheimer: So to be clear we need the following variables:
* `CMAKE_INSTALL_SYSCONFDIR`: global fastfetch.conf goes here, only used for parsing the own config file
* `TARGET_DIR_ETC`: global config dir, used to parse config files of other applications.

Configuring should be done the following way:
* If both are given: fine
* If one is given, initialize the other with the same value
* If none is given, initialize `TARGET_DIR_ETC` with `${TARGET_DIR_ROOT}/etc` and `CMAKE_INSTALL_SYSCONFDIR` using `include(GNUInstallDirs)`.

LinusDierheimer: Note the updated comment.

CarterLi: Can you please do the code change? I'm not an expert of cmake.

LinusDierheimer: ofc.

PR #278, opened 2022-10-04
Fix macOS compile failure

draumaz:
Appears to be a simple typo.


LinusDierheimer:
This is unneeded one #274 is merged, but thanks anyway.

PR #282, opened 2022-10-09
Introduce Windows support

CarterLi:
This PR port fastfetch to Windows based on MSYS2 posix compatible layer. While I left the common code ( utsname, passwd, etc. ) unchanged, most module code has been reimplemented in pure Win32 API ( mostly with WMI )

It can be run inside of POSIX environment ( msys bash ) or be run directly. When running inside bash, I reused Linux code to detect Linux package managers and shells; when running outside, Win32 APIs were used.

Running inside msys bash with msys2's default terminal mintty ( with Linux environment, `/proc/pid/*` exists ):

image

Running inside PowerShell with Windows Terminal

image

Running inside CMD with conhost

image

Prebuilt binaries: https://github.com/CarterLi/fastfetch/releases

Because of the huge code change, I hope it gets merged before another conflicts introduced


CarterLi:
I think the Windows support is good enough to have a release. Please review @LinusDierheimer

CMakeLists.txt
```diff
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 3.12.0) # target_link_libraries with OBJECT libs
 
 project(fastfetch
     VERSION 1.7.2
-    LANGUAGES C
+    LANGUAGES C CXX # Windows part requires C++ compiler
```
LinusDierheimer, line 5: I would prefer to call `enable_language(CXX)` only on windows.

CMakeLists.txt
```diff
@@ -13,10 +13,12 @@ project(fastfetch
 
 if("${CMAKE_SYSTEM_NAME}" MATCHES ".*[Ll]inux.*")
     set(LINUX TRUE CACHE BOOL "..." FORCE) # LINUX means GNU/Linux, not just the kernel
+elseif("${CMAKE_SYSTEM_NAME}" MATCHES "MSYS|Windows")
+    set(WIN_MSYS TRUE CACHE BOOL "..." FORCE) # Windows on msys2
```
LinusDierheimer, line 17: Cmake enables a `WIN32` variable on Windows by default, can't we used that one?

If this patch doesn't work without MSys2, cmake also enables a `MSYS` variable there.

See https://gitlab.kitware.com/cmake/community/-/wikis/doc/tutorials/How-To-Write-Platform-Checks#platform-variables

CMakeLists.txt
```diff
@@ -297,6 +322,44 @@ if(LINUX OR BSD)
...
+        src/detection/packages/packages_windows.c
+        src/detection/kernel/kernel_windows.cpp
+        src/detection/localip/localip_windows.c
+        src/util/windows/wmi.cpp
+
+        # Shared
+        src/detection/terminalfont/terminalfont_linux.c
+        src/detection/poweradapter/poweradapter_linux.c
+        src/detection/media/media_linux.c
+        src/detection/temps/temps_linux.c
```
LinusDierheimer, line 362: I would prefer to have "Not implemented" methods for those, since all of them won't work on windows.

CMakeLists.txt
```diff
@@ -417,6 +486,16 @@ if(APPLE)
...
+elseif(WIN_MSYS)
+    target_link_libraries(libfastfetch
+        PRIVATE "-lwbemuuid"
+        PRIVATE "-lole32"
+        PRIVATE "-loleaut32"
+        PRIVATE "-ldwmapi"
+        PRIVATE "-lopengl32"
+        PRIVATE "-lgdi32"
+        PRIVATE "-liphlpapi"
+    )
```
LinusDierheimer, line 498: Please remove the "-l" prefix

LinusDierheimer (commented):
Wow, thats a lot of work done. I think the best way to test this is just merge it into master and test it for a few days before i'll do the release.

CMakeLists.txt
CarterLi: Didn't know that

CarterLi:
Fixed. I also merged some *unimplemented* files into one

CarterLi:
> Wow, thats a lot of work done.

Better to add a Windows action, but I don't know if msys is supported.

> I think the best way to test this is just merge it into master and test it for a few days before i'll do the release.

We may release macOS first, as the current version still crashes on Intel

LinusDierheimer:
I'll do a point release then for all the new MacOS features, and maybe we can get Windows Support without MSys for 1.8

PR #286, opened 2022-10-11
Improvements, see commit list

CarterLi:


.github/workflows/push.yml
```diff
@@ -123,6 +123,71 @@ jobs:
...
+      # - name: run flashfetch
+      #   run: ./flashfetch
+
+      - name: upload artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: fastfetch-windows
+          path: |
+            ./*.dll
+            ./*.exe
```
LinusDierheimer, line 189: I think we should update CPack to build an windows installer and provide that instead of raw exe + dll.

CarterLi: I'm not familier with CPack, but I don't think fastfatch should be *installed* in `Program Files` or somewhere because it's not where CUI programs should be and it's not in global $PATH.

Same case of macOS, we should not provide .dmg because files in .dmg usually means that they should be dropped into `/Applications`. CUI tools like fastfetch should always be installed with package manager or just be put into somewhere in user's home folder.

LinusDierheimer: I am not familiar with the install process on those platforms, but this means we should remove the dmg.

LinusDierheimer: Can we at least zip the exe and the dlls together?

CarterLi: They will be zipped together by `actions/upload-artifact@v3`

Note Windows part doesn't go to release action (yet)

PR #291, opened 2022-10-12
Windows: improve Win32 API compatibility

CarterLi:


CMakeLists.txt
```diff
@@ -76,6 +77,15 @@ if(NOT CMAKE_BUILD_TYPE)
 endif()
 
 message(STATUS "Build type: ${CMAKE_BUILD_TYPE}")
+if(ENABLE_THREADS)
+    if(CMAKE_USE_PTHREADS_INIT)
+        message(STATUS "Threads type: pthread")
+    else()
+        message(STATUS "Threads type: Win32 thread")
+    endif()
```
LinusDierheimer, line 85:
```cmake
message(STATUS "Threads type: ${CMAKE_THREAD_LIBS_INIT}")
```

CMakeLists.txt
```diff
@@ -482,6 +492,14 @@ ff_lib_enable(OPENCL OpenCL)
 ff_lib_enable(LIBCJSON libcjson)
 ff_lib_enable(FREETYPE freetype2)
 
+if(ENABLE_THREADS)
+    target_compile_definitions(libfastfetch PRIVATE FF_HAVE_THREADS)
+
+    if(CMAKE_USE_PTHREADS_INIT)
+        target_link_libraries(libfastfetch PRIVATE Threads::Threads)
+    endif()
```
LinusDierheimer, line 500: I don't think we need the if, `Threads::Threads` will be empty if threads is provided by the standard lib.

LinusDierheimer:
I like the abstraction away from pthreads!

CMakeLists.txt
CarterLi: Just tested it. CMAKE_THREAD_LIBS_INIT was empty on my mac

LinusDierheimer: Ah that may be because pthread is part of libc there.

CarterLi: Let's keep it as it is.

PR #293, opened 2022-10-13
Support Native Windows

CarterLi:


CMakeLists.txt
```diff
@@ -15,20 +15,24 @@ if("${CMAKE_SYSTEM_NAME}" MATCHES ".*[Ll]inux.*")
...
 #############################
 
-set(THREADS_PREFER_PTHREAD_FLAG ON)
-find_package(Threads)
+if(WIN32)
+    set(Threads_FOUND ON) # Force using Win32 Threads on Windows, even if winpthread is found
+else()
+    set(THREADS_PREFER_PTHREAD_FLAG ON)
+    find_package(Threads)
+endif()
```
LinusDierheimer, line 35: Not against the current implementation, but i think it would be cleaner to prefer pthread only on linux. It is recommended there by the [doc](https://cmake.org/cmake/help/latest/module/FindThreads.html#variables-affecting-behavior) but unneeded on other platforms.

```cmake
set(THREADS_PREFER_PTHREAD_FLAG ${NOT WIN32})
find_package(Threads)
```

src/common/library.h
```diff
@@ -3,7 +3,16 @@
 #ifndef FF_INCLUDED_common_library
 #define FF_INCLUDED_common_library
 
-#include <dlfcn.h>
+#if defined(_WIN32) //We don't force MSYS to use LoadLibrary because dlopen also searches $LD_LIBRARY_PATH
+    #include <libloaderapi.h>
+    // Use #define to avoid `RTLD_LAZY is not defined` errors
+    #define dlopen(path, mode) ((void*)LoadLibraryA((path)))
+    #define dlsym(handle, symbol) ((void*)GetProcAddress((HMODULE)(handle), (symbol)))
```
LinusDierheimer, line 10: I would prefer to have them as static inline methods, we can just define `RTDL_LAZY` to anything and FF_UNUSED it in the method.

src/detection/uptime/uptime.h
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+#ifndef FF_INCLUDED_detection_uptime_uptime
+#define FF_INCLUDED_detection_uptime_uptime
+
+#include "fastfetch.h"
+
+uint64_t ffDetectUptime(FFinstance* instance);
```
LinusDierheimer, line 8: instance should be const.

LinusDierheimer (commented):
Good work!

LinusDierheimer:
And the CI has errors, flashfetch segfaults in MSys2.

PR #295, opened 2022-10-14
Remove msys target and other improvements

CarterLi:


LinusDierheimer:
I don't see anything wrong here, thanks!

PR #297, opened 2022-10-14
Update information on different platforms

CarterLi:
My English is not good. Feel free to correct me if anything is not right


LinusDierheimer:
My english isn't really advanced either, but it's fine i guess.

Issue #294, opened 2022-10-13
Support bar output for percentage format


CarterLi:
From winfetch:

![image](https://user-images.githubusercontent.com/6134068/195650667-f09fd8e1-ceda-419f-b176-5195fea61eed.png)

Looks very good to me

LinusDierheimer:
Closed with #308.

PR #308, opened 2022-10-18
Global: support bar output for percentage value #294

CarterLi:
image

We may add some options to control whether bar or number or both bar and number should be displayed.
Please tell me if you like this

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -221,6 +221,7 @@ set(LIBFASTFETCH_SRC
     src/common/settings.c
     src/common/library.c
     src/common/networking.c
+    src/common/bar.c
     src/logo/logo.c
     src/logo/builtin.c
     src/logo/image/image.c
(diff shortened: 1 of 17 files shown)


Samisafool:
What do the different colors in disk mean

src/common/printing.c
```diff
@@ -163,3 +163,33 @@ void ffPrintUserString(const char* value)
         ++value;
     }
 }
+
+void ffPrintPercentBar(uint8_t percent, uint8_t green, uint8_t yellow, uint8_t red)
```
LinusDierheimer, line 167: Maybe add an assert / documentation that these numbers should add up to 100. I don't think this is intuitive.

CarterLi:
> While i won't use it personally, i can clearly see why people might want this, and therefore want to merge it. But can we have a flag to turn it on?

Done

PR #309, opened 2022-10-18
Swap: split code out from memory detection

CarterLi:
1. Since swap module is disabled by default, swap detection is unnecessary in most cases.
2. On platform other than Linux, swap detection doesn't share code with memory.
3. Swap detection can be expensive. On Windows, it requires another WMI query which can lag heavily in some cases. Don't pay for you don't use.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -235,6 +235,7 @@ set(LIBFASTFETCH_SRC
     src/detection/cpuUsage/cpuUsage.c
     src/detection/gpu/gpu.c
     src/detection/memory/memory.c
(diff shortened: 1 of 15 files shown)


LinusDierheimer:
I agree that is the better way for now.

For the future, i am working on the format function taking callbacks instead of values. That plus the config system that will allow the same module printed multiple times with different configurations, will allow us to merge a lot of modules back together (the best example is `Time`, `Date` and `DateTime`)

LinusDierheimer:
https://github.com/LinusDierheimer/fastfetch/discussions/310 Thats what the last point is for.

PR #312, opened 2022-10-20
Android: add vulkan support for GPU detection

CarterLi:
Termux does have package pciutils, but it's put inside of root-repo and and doesn't provide libpci.pc. So I think use vulkan can be a better option than libpci because vulkan.so is provided by Android system.

To build fastfetch with android support, install `vulkan-headers` and `vulkan-loader-android` (for `vulkan.pc`). Termux package needs updating.


LinusDierheimer:
Termux package needs to be updated for 1.8 anyway, because of the freetype dependency.

PR #313, opened 2022-10-22
Fixes and performance improvements

CarterLi:


src/common/printing.h
```diff
@@ -12,7 +12,20 @@ void ffPrintFormat(FFinstance* instance, const char* moduleName, uint8_t moduleI
...
+    // However, because the application buffer are shared between threads ( since stdout itself is shared between threads ),
+    // every fstdio function call requires exclusive buffer locking to ensure thread safety
+    // which means
+    // putchar(c);
+    // is equivalent to
+    // flockfile(stdout); putchar_unlocked(c); funlockfile(stdout);
+    // Since putchar is a simple buffer appending operation before the buffer is full, putchar(c) can be hundred times slower than putchar_unlocked
+    // To improve performance, it's always prefered to fill a buffer then fwrite rather than putchar loop
+    // see also flockfile(3)
+    fwrite(&c, 1, times, stdout);
```
LinusDierheimer, line 27: This will write c, but then the next times - 1 random bytes that happen to be after c in memory.

If we want to use a buffer, it would need to be done the following time:
```c
char* buf = malloc(times);
memset(buf, c, times);
fwrite(buf, sizeof(char), times, stdout);
free(buf);
```
But indeed probably still faster than the current implementation.

CarterLi:
We can actually `#define putchar putchar_unlocked` when multithreading is disabled. most fstdio functions have fstdio_unlocked except printf and scanf

PR #315, opened 2022-10-23
Packages: support Chocolatey (Windows)

CarterLi:


CarterLi:
What about a new release for Windows? I have used fastfetch personally for several weeks and have fixed all issues I found.

I think it's stable enough (at least for Windows 11) to release a new version, so package managers can collect fastfetch.

LinusDierheimer:
Thanks as always.

PR #324, opened 2022-10-29
Logo: add Parabola GNU/Linux-libre logo

jumps-are-op:
Ref: https://github.com/dylanaraps/neofetch/blob/master/neofetch#L9360-L9392

Parabola is a fully free GNU/Linux operating system https://parabola.nu/

PARABOLA fastfetch


LinusDierheimer:
Can you please open a PR for only the logo? I want to solve the issue in a different way to keep consistency with other invalid indicies.

jumps-are-op:
Oh, I think I forgot to make a separate branch for the logo.

PR #327, opened 2022-11-01
Logo: Add Parabola GNU/Linux-libre logo (instead of #324)

jumps-are-op:
Ref: https://github.com/dylanaraps/neofetch/blob/master/neofetch#L9360-L9392

Parabola is a fully free GNU/Linux operating system https://parabola.nu/

PARABOLA fastfetch


LinusDierheimer:
Thank you!

PR #328, opened 2022-11-01
Fix setting `cpu-format`

jumps-are-op:
`--cpu-format` set `terminal-format` instead of `cpu-format` because of a typo in fastfetch.c


LinusDierheimer:
Oh thanks.

PR #332, opened 2022-11-06
Various improvements

CarterLi:
image

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,6 +7,8 @@ project(fastfetch
     HOMEPAGE_URL "https://github.com/LinusDierheimer/fastfetch"
 )
 
+set(PROJECT_LICENSE "MIT license")
+
 ###################
 # Target Platform #
 ###################
@@ -577,6 +579,26 @@ add_executable(flashfetch
 target_link_libraries(flashfetch
     PRIVATE libfastfetch
 )
+
+if(WIN32)
+    if(PROJECT_VERSION_TWEAK)
+        string(REGEX MATCH "[0-9]+" PROJECT_VERSION_TWEAK_NUM "${PROJECT_VERSION_TWEAK}")
+    else()
+        set(PROJECT_VERSION_TWEAK_NUM "0")
+    endif()
+
+    set(TARGET_NAME fastfetch)
+    configure_file(src/util/windows/version.rc.in version.fastfetch.rc)
+    target_sources(fastfetch
+        PRIVATE version.fastfetch.rc
+    )
+
+    set(TARGET_NAME flashfetch)
+    configure_file(src/util/windows/version.rc.in version.flashfetch.rc)
+    target_sources(flashfetch
+        PRIVATE version.flashfetch.rc
+    )
+endif()
 
 ###################
 # Testing targets #
--- a/src/common/init.c
+++ b/src/common/init.c
@@ -125,7 +125,6 @@ static void initState(FFstate* state)
     #ifdef WIN32
     //https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/setlocale-wsetlocale?source=recommendations&view=msvc-170#utf-8-support
     setlocale(LC_ALL, ".UTF8");
-    setvbuf(stdout, NULL, _IOFBF, 4096);
     #endif
 
     state->logoWidth = 0;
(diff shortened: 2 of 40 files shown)


CMakeLists.txt
```diff
@@ -570,6 +570,11 @@ add_executable(fastfetch
 target_link_libraries(fastfetch
     PRIVATE libfastfetch
 )
+if(WIN32)
+    target_sources(fastfetch
+    PRIVATE src/util/windows/version.rc
+)
```
LinusDierheimer, line 576: Doesn't flashfetch need that too?

src/util/windows/version.rc
```diff
@@ -0,0 +1,49 @@
...
+			VALUE "Comments",         "https://github.com/LinusDierheimer/fastfetch"
+			VALUE "CompanyName",      "Github"
+			VALUE "FileDescription",  "Like neofetch, but much faster because written in C."
+			//TODO: Generate automatically with git tag
+			VALUE "FileVersion",      "V1.7.5.0"
+			VALUE "InternalName",     "fastfetch.exe"
+			VALUE "LegalCopyright",   "MIT license"
+			VALUE "OriginalFilename", "fastfetch.exe"
+			VALUE "ProductName",      "fastfetch"
+			VALUE "ProductVersion",	  "V1.7.5.0"
```
LinusDierheimer, line 41: Can we configure this (especially the version) from cmake using `configure_file`?

CarterLi: I have no idea how to do that.

CarterLi:
Done

src/common/init.c
```diff
@@ -337,6 +336,7 @@ BOOL WINAPI consoleHandler(DWORD signal)
     if(signal == CTRL_C_EVENT)
     {
         resetConsole();
+        exit(0);
         return TRUE;
     }
```
jumps-are-op, line 341: `return TRUE;` will never be executed because of `exit(0);`

src/detection/disk/disk_linux.c
```diff
@@ -73,9 +73,9 @@ void ffDetectDisksImpl(FFDiskResult* disks)
         #endif
 
         //Detects stats
-        struct statvfs fs;
-        if(statvfs(disk->mountpoint.chars, &fs) != 0)
-            memset(&fs, 0, sizeof(struct statvfs)); //Set all values to 0, so our values get initialized to 0 too
+        struct statvfs64 fs;
+        if(statvfs64(disk->mountpoint.chars, &fs) != 0)
+            memset(&fs, 0, sizeof(struct statvfs64)); //Set all values to 0, so our values get initialized to 0 too
```
LinusDierheimer, line 78: I would put a
```c
#ifndef _LARGE_FILE_API
    #define _LARGE_FILE_API
#endif
```
at the top of the file to ensure the prototype is available on all platforms.

src/modules/gpu.c
```diff
@@ -44,12 +44,6 @@ void ffPrintGPU(FFinstance* instance)
     if(ffPrintFromCache(instance, FF_GPU_MODULE_NAME, &instance->config.gpu, FF_GPU_NUM_FORMAT_ARGS))
         return;
 
-    if(ffStrbufCompS(&ffDetectHost()->productName, FF_HOST_PRODUCT_NAME_WSL) == 0)
-    {
-        ffPrintError(instance, FF_GPU_MODULE_NAME, 0, &instance->config.gpu, "WSL doesn't expose senseful GPU names");
-        return;
-    }
-
```
LinusDierheimer, line 52: I don't see how GPU is detected on WSL other then pci, which will always expose the same name on all systems. Why delete this?

CarterLi: Because that's what reported by the system. WSL2 is a virtual machine like vmware or virtualbox, which only provides virtual GPUs. As we don't detect specially for other VMs, we shouldn't detect WSL either. In addition, As WSL is continually improving, WSL may report the real GPU names some day. If a user don't like it, just disable the module with `--structure`

Note: we  can actually detect the real GPU names with [wmic.exe](https://learn.microsoft.com/en-us/windows/win32/wmisdk/wmic). However since we are running in WSL, we should honor what it reports.

src/detection/disk/disk_linux.c
CarterLi: Well, should it be defined by glibc? If it's not defined, it might be disabled when configuring.

LinusDierheimer: Ah right. We might switch between `statvfs64` and `statvfs` at compile time, depending weather its defined or not.

CarterLi: Ok

LinusDierheimer:
Thanks as always!

PR #335, opened 2022-11-12
Windows / macOS: simplify code

CarterLi:
1. simplify resource management by using C++ RAII like GCC attribute extension `cleanup`
2. simplify HEKY reading

Actually `__attribute__((__cleanup__(cleanupFn)))` is supported on all major C compiler ( gcc, clang and icc ) except MSVC I only used it on Windows and macOS for now. Let's see if the project author like it.

### DIFF
--- a/src/common/networking_windows.c
+++ b/src/common/networking_windows.c
@@ -75,7 +75,7 @@ bool ffNetworkingSendHttpRequest(FFNetworkingState* state, const char* host, con
         }
     }
 
-    FFstrbuf command;
+    FF_STRBUF_AUTO_DESTROY command;
     ffStrbufInitA(&command, 64);
     ffStrbufAppendS(&command, "GET ");
     ffStrbufAppendS(&command, path);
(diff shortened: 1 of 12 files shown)


CarterLi:
[__attribute__((__cleanup__(cleanupFn)))](https://gcc.gnu.org/onlinedocs/gcc-12.2.0/gcc/Common-Variable-Attributes.html#Common-Variable-Attributes) runs a function when the variable goes out of scope.

So following code

```c
FFstrbuf str;
ffStrbufInit(&str, "some string");

if(condition)
{
    ffStrbufDestroy(&str);
    return false;
}

//Use str
ffStrbufDestroy(&str);
return true;
```

Can be simplified to

```c
FFstrbuf __attribute__((__cleanup__(ffStrbufDestroy))) str;
ffStrbufInit(&str, "some string");

if(condition)
    return false;

//Use str
return true;
```

CarterLi:
Merged to my dev branch to avoid future conflicts

PR #352, opened 2022-12-17
Global: remove caching functions except image caching

CarterLi:
Motivation:

1. Caching doesn't gain too much performance boost. According to my benchmark, `fastfetch` without caching is only about  10ms faster than `fastfetch --nocache`. Most truly expensive operations such as networking can't be cached.

image

2. We don't have a strategy to invalidate caching.  Most users newly come to fastfetch are not aware of caching, resulting in issues https://github.com/LinusDierheimer/fastfetch/issues/346

3. Caching increases code complexity. We often need different branches to handle cached results

4. There are bugs affects caching: https://github.com/LinusDierheimer/fastfetch/issues/342

5. Caching is annoying when debugging. I have experienced several times that the result never changes no matter how I modify the code. It's a small issue, but does decrease coding experience.

I left this PR for discussion. If it gets no attentions, I will merge it.


LinusDierheimer:
I am fine with merging it. Caching was required when detection wasn't as good, but now it indeed doesn't really boost performance anymore.

LinusDierheimer (approved):

Issue #354, opened 2022-12-21
Update /src/data/help.txt


kojq:
`fastfetch --version`: `1.7.5` and 225d1dd

When doing `fastfetch -h`, output includes:
`--color-keys <color>:        sers the color of the keys`

**sers** should be **sets**.

kojq:
Other changes were made in the PR.

PR #355, opened 2022-12-21
Update help.txt

kojq:
Also, consider changing `--disable-linewrap` to `--disable-line-wrap`.


kojq:
This resolves #354

kojq:
I did some cleanup. Also, gotta remove this if #352 comes through:
```
    -r,--recache <?value>:            Generate new cached values
    --nocache <?value>:               Don't use cached values, but also don't overwrite existing ones
```

kojq:
For binary choices, the word choice is inconsistent. Is `whether` or `set if` preferred?

kojq:
Superfluous and ambiguous language have been removed.

kojq:
Some different mistakes are in `/src/data/config_user.txt`. Should that file be changed separately because it is a bit long and shares some information with `/src/data/help.txt` as well?

LinusDierheimer:
Thanks for your work on this. Feel free to add other help files to this PR too!

> For binary choices, the word choice is inconsistent.

Choose what sounds better to you. I'm not a native speaker, so i can' really tell.

PR #357, opened 2022-12-21
Added OpenMandriva

AloneER0:
image
There shouldn't be any issues, but if there are inform me about them and i will try to resolve them as fast as possible.


LinusDierheimer:
Thanks

PR #363, opened 2022-12-29
Bump CMakeList version

CarterLi:
also try supporting bios module on freebsd


CarterLi:
For some reaason Windows binaries were not packaged into a zip file. Can you have a look?

PR #378, opened 2023-01-01
fixed formatting error

IamNoRobot:
Hi, 
sorry, there was a formatting error, resulting in an error in compiling that I overseen (as it didn't cause trouble). Fixed it tough, no errors now!
Sorry for that and thanks for adding to the main branch!


LinusDierheimer:
No problem, i really should have waited for the CI to pass.

PR #389, opened 2023-01-13
RFC: an attempt to add a new module which can run custom shell scripts

CarterLi:
Unlike other modules, `Command` module was designed to be able to be used mulitple times. A special option `--command-text` can be set multiple times; each option specifies one script to be executed.

For example:

```
$ ./fastfetch -s title:separator:os:command:command:command \
            --command-key GCC --command-text 'gcc --version | head -1 | awk \'{print $3}\'' \
            --command-key CMake --command-text 'cmake --version | head -1 | awk \'{print $3}\'' \
            --command-key PkgConfig --command-text 'pkg-config --version'
```

image

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -263,6 +263,7 @@ set(LIBFASTFETCH_SRC
     src/modules/cpuUsage.c
     src/modules/cursor.c
     src/modules/custom.c
+    src/modules/command.c
     src/modules/date.c
     src/modules/datetime.c
     src/modules/de.c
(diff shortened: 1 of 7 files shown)


LinusDierheimer:
I still think we should add a general way to print a module multiple times with different configs. A lot of modules would benefit from that.

But for now i think this MR is fine.

CarterLi:
Yeah. Apart from https://github.com/LinusDierheimer/fastfetch/discussions/310, I have a thought that embed a script language such as lua, so that users have full control of module formatting and printing.

But those are all the long-time plan. For now this PR is the easiest solution I can come up with.

LinusDierheimer (approved):

PR #399, opened 2023-01-21
Release 1.9.0

CarterLi:
Today (UTC+8) is the biggest festival in my country [Chinese new year](https://en.wikipedia.org/wiki/Chinese_New_Year). I'd like to cut a release to celebrate it.

I think the number of new features are big enough to increase the minor version (`major.minor.patch`). What do you think @LinusDierheimer


LinusDierheimer:
Yeah, lets do it.

PR #406, opened 2023-01-29
Preserve 4 significant gigits when parsing size

Nefsen402:
This algorithm will try to preserve at most 4 significant digits. It will also try to drop unnecessary trailing zeros.


CarterLi:
Motivition?

Nefsen402:
> Motivition?

I'd like to see more than just "3 TiB / 4TiB" for my disks.

LinusDierheimer:
I'd prefer to have this as an option, maybe `--binary-digits` (similar to `--binary-prefix`) and let it default to 0, which means keep the current behavior.

luni3359:
Even if it's a couple more characters I think it's starting to look too crowded imo

CarterLi:
Not planned

PR #411, opened 2023-02-01
fix: add another mutter detection for linux mint 21.1 (cinnamon)

Zerogiven:
On Linux Mint 21.1 (cinnamon) i always got "Unknown WM: Mutter (Muffin)". Just added another comparision now i'm able to get the correct WM theme.

### DIFF
--- a/src/detection/displayserver/linux/wmde.c
+++ b/src/detection/displayserver/linux/wmde.c
@@ -70,7 +70,8 @@ static void applyPrettyNameIfWM(FFDisplayServerResult* result, const char* proce
         strcasecmp(processName, "gnome-shell") == 0 ||
         strcasecmp(processName, "gnome shell") == 0 ||
         strcasecmp(processName, "gnome-session-binary") == 0 ||
-        strcasecmp(processName, "Mutter") == 0
+        strcasecmp(processName, "Mutter") == 0 ||
+        strcasecmp(processName, "Mutter (Muffin)") == 0
     ) ffStrbufSetS(&result->wmPrettyName, FF_WM_PRETTY_MUTTER);
     else if(
         strcasecmp(processName, "cinnamon-session") == 0 ||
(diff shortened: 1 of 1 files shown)


src/detection/displayserver/linux/wmde.c
```diff
@@ -70,7 +70,8 @@ static void applyPrettyNameIfWM(FFDisplayServerResult* result, const char* proce
         strcasecmp(processName, "gnome-shell") == 0 ||
         strcasecmp(processName, "gnome shell") == 0 ||
         strcasecmp(processName, "gnome-session-binary") == 0 ||
-        strcasecmp(processName, "Mutter") == 0
+        strcasecmp(processName, "Mutter") == 0 ||
+        strcasecmp(processName, "Mutter (Muffin)") == 0
```
LinusDierheimer, line 74: This should go one below to the detection of `Muffin`, not `Mutter`.

On another note, where does that value come from? Doesn't seem to be either a process name nor an environment variable.

LinusDierheimer (changes requested):

LinusDierheimer:
Is it possible that you wanted to change something here: https://github.com/LinusDierheimer/fastfetch/blob/78ca4e7c87f04141e2dfeb1b8091dd6a2b4f415b/src/detection/wmtheme/wmtheme_linux.c#L209
This is where the wm theme module looks at which wm is running

LinusDierheimer (approved):

src/detection/displayserver/linux/wmde.c
Zerogiven: I see... moved it down to Muffin detetection :+1: But to be honest: last time i dived into C is about 10 years ago :D However, i'm also pretty sure that this i not the process name, but i did not look where you set the processName var :)

LinusDierheimer: `processName` is a bit misleading to be honest. Originally i just went through all running processes and checked their names, thus the name. Now it can be either (in that order checked):
* The name the WM reports itself as when connecting to it
* A value set in one of the many standard env variables
* A process name of any running process

Maybe `Mutter (Muffin)` is coming from the first point. You can run `xprop -id "${"$(xprop -root -notype _NET_SUPPORTING_WM_CHECK)"##* }" -notype -len 100 -f _NET_WM_NAME 8t` to check for that.

LinusDierheimer: Great. I'll merge as soon as the pipeline succeeds.

Zerogiven: Just another side note: wmctrl returns also Mutter (Muffin)

```bash
❯ wmctrl -m
Name: Mutter (Muffin)
Class: N/A
PID: N/A
Window manager's "showing the desktop" mode: OFF
```

PR #412, opened 2023-02-01
fix: spell correction

Zerogiven:
Just cause i'm sometimes a bit fussy ;D
https://capitalizemytitle.com/successful-vs-successfull-what-is-the-right-spelling/


LinusDierheimer:
Thanks.

PR #425, opened 2023-02-22
sound_apple.c: handle `kAudioObjectPropertyElementMain` for macOS < 12

nandahkrishna:
We were having issues with building the latest `fastfetch` on macOS < 12 at Homebrew (see Homebrew/homebrew-core#123899) due to the use of `kAudioObjectPropertyElementMain` in `sound_apple.c`. This is not present in older versions of macOS as it was renamed from `kAudioObjectPropertyElementMaster` in macOS 12.

This PR defines a macro to solve the issue on macOS < 12, and these changes have been ~inspired by~ stolen from those made in https://github.com/libsdl-org/SDL/commit/b0dc6709b9735d22b2713a7b76e5bc0892a0e5df.

### DIFF
--- a/src/detection/sound/sound_apple.c
+++ b/src/detection/sound/sound_apple.c
@@ -2,6 +2,11 @@
 #include "util/apple/cf_helpers.h"
 
 #include <CoreAudio/CoreAudio.h>
+#include <AvailabilityMacros.h>
+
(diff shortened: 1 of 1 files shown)


CarterLi:
Thanks.

PR #432, opened 2023-03-09
[Ready to merge] Initial support of JSON config file, the first step to fastfetch 2.0

CarterLi:
This PR introduces a JSON config file support for fastfetch. The reason JSON (actually, `JSONC`, or `JSON with comments`) was chosen instead of YAML was that I could not find an easy-to-use C yaml parser library. I did try libyaml, and found it basically gave me a pile of tokens. I gave up.

Currently JSON config file support was implemented as a special module `JsonConfig`, and can be tested with `fastfetch -s jsonconfig`. This is a temporary solution and should be discussed further.

Detection code was not touched yet. We have many singleton detection results, and they don't work with different config well. Further discussion needed.

Currently only a few modules were supported by `JsonConfig`, while old path ( `--structure` ) were not affected.

Demo: Following JSONC document

```jsonc
// ~/.config/fastfetch/config.jsonc
{
  "modules": [
    {
      "type": "title",
      "fdqn": true // Affect separator
    },
    "separator", // Same as { "type": "separator" }
    "os", // Same as { "type": "os" }
    {
      "type": "battery",
      "temp": true // This option is used in detection code, doesn't work yet
    },
    {
      "type": "command",
      "key": "Compiler",
      "text": "cc -v 2>&1 | head -1"
    },
    {
      "type": "command", // JSON config is designed for printing the same module multiple times
      "key": "Editor",
      "text": "$EDITOR -v | head -1"
    },
    "datetime",
    {
      "type": "datetime",
      "key": "Date",
      "format": "{1}-{3}-{11}"
    },
    {
      "type": "datetime",
      "key": "Time",
      "format": "{14}:{17}:{20}"
    }
  ]
}
```

generates following output

image


EshaqqMi:
I'm impressed 😁 with ur work, and **MOTIVATED**

LinusDierheimer:
Looking good so far!

When we merge this, i will move fastfetch to an organization, because it has really outgrown being a personal project.

tdtooke:
Makes since.  At work JSON is my go to for this sort of thing.

CarterLi:
It has been 3 months, but I finally did it. JSON config support has been added to All 51 modules.

As JSON config is designed for printing modules multiple times, some old features implemented for this function has been obseleted:

1. `--set` and `--set-keyless` have been replaced to Module `Custom`, which is used for printing custom strings, with or without keys.
2. Modules `Date` and `Time` have been removed. Use `DateTime` instead.
3. Module `Command` is reworked, and behaves like a normal module.

PR #435, opened 2023-03-18
Small update for nobara logo

regulargvy13:
The last line in nobara logo was missing just a few spaces, so I added them, and now it's in proper shape :)


CarterLi:
Thanks

PR #445, opened 2023-04-10
Update Windows 11 ASCII logo to look more visually consistent.

KingKrouch:
Reason being that the original ASCII art as shown in Windows Terminal has random portions using "*" and "(" as opposed to "/". The Windows 10 and Windows 10 small ASCII icons don't have this problem, so I wanted to make it look more consistent.


LinusDierheimer:
Thanks. I like it.

PR #446, opened 2023-04-12
Adds another font color index to arch icon

tdtooke:
Now the Arch icon has 2 font color indexes.  1 is the top, 2 is the bottom which will allow for customization so old neofetch themes like ozozfetch are possible.  By default both font colors are the same so the user has to specify if he/she wants to change them for a theme.


LinusDierheimer:
Thank you!

PR #463, opened 2023-05-28
perf improvement attempts for OpenGL.c

hilga007:
avoiding unnecessary memory allocations and string copying

### DIFF
--- a/src/modules/opengl.c
+++ b/src/modules/opengl.c
@@ -7,36 +7,29 @@
 
 void ffPrintOpenGL(FFinstance* instance)
 {
-    FFOpenGLResult result;
-    ffStrbufInit(&result.version);
-    ffStrbufInit(&result.renderer);
-    ffStrbufInit(&result.vendor);
-    ffStrbufInit(&result.slv);
+    FFOpenGLResult result = { 0 };
 
     const char* error = ffDetectOpenGL(instance, &result);
-    if(error)
+    if (error)
     {
         ffPrintError(instance, FF_OPENGL_MODULE_NAME, 0, &instance->config.openGL, "%s", error);
         return;
     }
 
-    if(instance->config.openGL.outputFormat.length == 0)
+    if (instance->config.openGL.outputFormat.length == 0)
     {
         ffPrintLogoAndKey(instance, FF_OPENGL_MODULE_NAME, 0, &instance->config.openGL.key);
         puts(result.version.chars);
     }
     else
     {
-        ffPrintFormat(instance, FF_OPENGL_MODULE_NAME, 0, &instance->config.openGL, FF_OPENGL_NUM_FORMAT_ARGS, (FFformatarg[]) {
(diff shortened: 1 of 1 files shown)


src/modules/opengl.c
```diff
@@ -7,36 +7,29 @@
 
 void ffPrintOpenGL(FFinstance* instance)
 {
-    FFOpenGLResult result;
-    ffStrbufInit(&result.version);
-    ffStrbufInit(&result.renderer);
-    ffStrbufInit(&result.vendor);
-    ffStrbufInit(&result.slv);
```
LinusDierheimer, line 14: If the strbufs aren't initialized, `ffDetectOpenGL` will write data in random memory places, which most probably will cause a seg fault

src/modules/opengl.c
```diff
@@ -7,36 +7,29 @@
...
+            { FF_FORMAT_ARG_TYPE_STRBUF, &result.renderer },
+            { FF_FORMAT_ARG_TYPE_STRBUF, &result.vendor },
+            { FF_FORMAT_ARG_TYPE_STRBUF, &result.slv }
+        };
 
-    ffStrbufDestroy(&result.version);
-    ffStrbufDestroy(&result.renderer);
-    ffStrbufDestroy(&result.vendor);
-    ffStrbufDestroy(&result.slv);
+        ffPrintFormat(instance, FF_OPENGL_MODULE_NAME, 0, &instance->config.openGL, FF_OPENGL_NUM_FORMAT_ARGS, formatArgs);
```
LinusDierheimer, line 33: Why? The compiled assembly will be the same

src/modules/opengl.c
LinusDierheimer: The reason that the pipeline doesn't fail is that the GitHub VMs don't have OpenGL support

src/modules/opengl.c
hilga007: Again - did not mean to submit to you - the "why" is just "this is what I have been learning w/o a class and w/o a university and wanted to apply a small change based on best practice" 

I guess so maybe what you are saying is that "it does not matter which path is taken, so long as the destination is reached?" Whereas that is the opposite of what I had bothered to learn, which is why this change had been made. 

"Conformance is performance" was one of the small sayings.

LinusDierheimer: The example is similar to the following:
```c
void do_something(int);

//My Version
do_something(5);

//Your version
int a = 5;
do_something(a);
```
Compiled they are the same, so lets only talk about the code style. Mine is more simple, and still readable, so it is the better solution here. (In my opionion, we are at a point where personal opionion does make the difference)

hilga007: And experience! Your experience makes it make sense! Thank you!!! As they say "real world experience" makes a difference over "textbook experience" :) :) 

Sorry for the panic attack..... 
Again I am very new with this. Only reading C99 books when young, and then 2 months of a course + textbook.

PR #465, opened 2023-05-28
tidy bios_linux.c

hilga007:
use logical negation (!) and replace the != true checks with || to be more concise and performant.

### DIFF
--- a/src/detection/bios/bios_linux.c
+++ b/src/detection/bios/bios_linux.c
@@ -3,38 +3,41 @@
 
 #include <stdlib.h>
 
-static bool hostValueSet(FFstrbuf* value)
+static bool hostValueSet(const FFstrbuf* value)
 {
-    return
-        value->length > 0 &&
-        ffStrbufStartsWithIgnCaseS(value, "To be filled") != true &&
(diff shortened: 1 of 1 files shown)


LinusDierheimer:
This doesn't make a difference.

hilga007:
OK so do you mind explaining why? Again this was an accidental submition to you, but I have only been learning solo from a book with web UI for test items. 

According to the learning activities - yes, to the authors of the book  it does make a difference. The textbook states that || is more concise and performant and is best practice.

LinusDierheimer:
An if stops evaluating as soon as the result is clear:
```c
//All must be true, so if b is false, c won't be evaluated, because the if will always be false
if(a && b && c) {}

//One must be true, so if a is false, and b is true, there is no reason to evaluate c anymore
if(a || b || c) {}
```
Usually you try to evaluate as little as possible, because that logically is more performant. What the better solution is depends on the situation, not the operator. Both are equally good.

Also readability is a factor. Code that can be easier read is usually better code.

LinusDierheimer:
In this example both are virtually the same. Both stop as soon as possible. You can show this with boolean algebra:
```c
// de morgan's laws
!a && !b && !c = !(a || b || c)
```
But i think my solution is more readable.

PR #496, opened 2023-07-12
CPU: properly detect CPU on POWER

pkubaj:
Excerpt for a thread from /proc/cpuinfo:
processor       : 0
cpu             : POWER9, altivec supported
clock           : 3800.000000MHz
revision        : 2.2 (pvr 004e 1202)

Without this patch, the following is printed for a CPU:
CPU: CPU (64) @ 3.8 GHz

With it, there is:
CPU: POWER9, altivec supported (64) @ 3.8 GHz


CarterLi:
Thanks

PR #508, opened 2023-08-01
Disk(Linux): fix byteUsed not correctly in ext4

xzl01:


CarterLi:
Thanks. I'll fix the BSD port

PR #515, opened 2023-08-08
Logo(Builtin): Fix `RebornOS`

shivanandvp:
This PR Fixes the RebornOS logo. My earlier logo is already a part of `fastfetch`, but this one improves it.

Screenshot from 2023-08-08 04-46-17


CarterLi:
Instead of dropping it, what about name the old logo rebornos_old?

shivanandvp:
> Instead of dropping it, what about name the old logo rebornos_old?

The actual logo has not changed since my first PR. This one just gets closer to the actual logo (better fidelity). I assumed it would bloat up your binary if the old code is not going to be used.

PR #534, opened 2023-08-23
Fix typo in config parsing for --color-title

IanManske:
Having `--color-title` set in the config currently causes fastfetch to error with `Error: unknown option: --color-title`.


CarterLi:
Thanks

PR #555, opened 2023-09-07
Fully updated evolutionOS Logos to my liking

Tracker-Friendly:
I like these ones better, and i fixed the various escape code issues (sorry!)


Tracker-Friendly:
By the way, thanks for your help on the other one, but the colors don't match up with the ones on neofetch, which is why i committed over it.

CarterLi:
You should at least remove those conflict markers.

Please, if you don't know how to make a PR, tell me what you want to change let me do it for you.

PR #560, opened 2023-09-17
Arguments: handle `--lib-pulse` properly

xbt573:
`--lib-pulse` is not handled properly because of `key` usage instead of `subkey`


CarterLi:
Thanks. I am superised that there are people actually using these options

PR #564, opened 2023-09-22
[WIP] Wayland: support fractional scale

LinusDierheimer:
WIP because:
* Kwin always reports a fractional scale of 1, despite the actual scale setting. Would be great if someone could test if that is a problem with my code or my compositor.
* There is only one global fractional scale factor, not one per monitor. At the moment i multiply the linear scale factor of each monitor with the fractional scale to get the scale per monitor. Not sure if this is optimal.

Despite that i refactored the code to be more how the libwayland authors wanted a client to use their lib.


CarterLi:
I had my monitor working in PIP mode, and made the big screen 2.5x and the small screen 1.5x

image

I tried mpv. When I move the mpv window between two screens, mpv can correctly detect the fractional scale factor.

image

However in fastfetch, it reports integral scale factor only.

image

CarterLi:
I'm pretty sure that the fractional scale factor value is tied to the screen of the wayland window is currently displaying on. So that you must have a window before you can detect the correct scale factor.

> At the moment i multiply the linear scale factor of each monitor with the fractional scale to get the scale per monitor. Not sure if this is optimal.

Probably no.

LinusDierheimer:
I think we can close this for now.

You need a shell to receive proper values for the monitor the shell is currently on. 2 Problems for fastfetch:
* Wayland forbids positioning of windows by the app for security reasons, so fastfetch can't place its shell on each monitor to get the associated fractional scale value. There are open merge requests for it, but no compositor implements it yet.
* Creating a shell, even one that is not presented (no surface attached to it), currently differes from compositor to compositor, because the standard is rapidly evolving (Most compositors advertise only one of `wl_shell`, `xdg_shell`, `zwlr_layer_shell_v1`, `org_kde_shell`, etc). Writing an implementation for each of them is not feasible. This will solve once `xdg_shell_v??` is properly adopted.

I think that for multi monitor setups, having fractional scale factor support is realistic only in a few years.

We can think about supporting it on single monitor setups with compositors that advertise `xdg_shell` tho.

PR #570, opened 2023-10-01
Fix Slackware colors

Alxhr0:
image

This is how the Slackware ascii looks like in neofetch, it's not fully blue.


CarterLi:
Thanks

PR #572, opened 2023-10-04
Fix several `--help` command related issues

iamrasel:


src/fastfetch.c
```diff
@@ -522,13 +524,13 @@ static inline void printCommandHelp(const char* command)
...
     {
-        constructAndPrintCommandHelpFormat("gamepad", "{1}", 1,
+        constructAndPrintCommandHelpFormat("gamepad", "{1}", 2,
             "Name",
             "Identifier"
         );
     }
     else
-        fprintf(stderr, "No specific help for command %s provided\n", command);
+        fprintf(stderr, "No specific help for command \'%s\' provided\n", command);
```
CarterLi, line 533: I don't think we need `\` here

src/fastfetch.c
```diff
@@ -439,7 +441,7 @@ static inline void printCommandHelp(const char* command)
             "Album name"
         );
     }
-    else if(ffStrEqualsIgnCase(command, "datetime-format") == 0 || ffStrEqualsIgnCase(command, "date-format") || ffStrEqualsIgnCase(command, "time-format"))
+    else if(ffStrEqualsIgnCase(command, "datetime-format") || ffStrEqualsIgnCase(command, "date-format") || ffStrEqualsIgnCase(command, "time-format"))
```
CarterLi, line 444: We can remove `|| ffStrEqualsIgnCase(command, "date-format") || ffStrEqualsIgnCase(command, "time-format")` as the modules `date` and `time` no longer exist.

CarterLi:
Thanks for the fix! Because they are not in module implementation files, I keep forgetting to update them.

I actually considered moving them into `module.c`, so that we can reuse the macros `FF_*_NUM_FORMAT_ARGS`, but I also forgot it.

iamrasel:
Just committed with suggested removals, please recheck.

PR #579, opened 2023-10-07
Logo: add Chimera Linux

JamiKettunen:
Logo and colors as seen in https://github.com/chimera-linux/cports/blob/master/contrib/neofetch/patches/chimera.patch

I have a question though (and will mark it as a draft for now): would it be possible to show `Chimera Linux` (similar to `neofetch`) instead of just `Chimera` on the `OS` line?

`cat /etc/os-release`:
```sh
NAME="Chimera"
ID="chimera"
PRETTY_NAME="Chimera Linux"
HOME_URL="https://chimera-linux.org"
DOCUMENTATION_URL="https://chimera-linux.org/docs"
LOGO="chimera-logo"
ANSI_COLOR="0;38;2;214;79;93"
```

`fastfetch` demo:
image

`neofetch` demo (for comparison):
image


CarterLi:
Fastfetch uses `NAME` instead of `PRETTY_NAME` to generate more constant and accurate results, because some distros include version or code name in `PRETTY_NAME` but some don't. Also see: https://github.com/dylanaraps/neofetch/issues/2381

The output can be customized though. You may try `fastfetch --os-format '{3}'` `fastfetch --help os-format`

JamiKettunen:
That indeed makes `OS: Chimera Linux` show up with `fastfetch --os-format '{3} {12}'` making it match `neofetch`; I suppose since for every other distribution it works properly it doesn't make sense to add in a special case where it would be defaulted to that for this particular one, though this also affects Void Linux with a similar `/etc/os-release`:
```
NAME="Void"
ID="void"
PRETTY_NAME="Void Linux"
HOME_URL="https://voidlinux.org/"
DOCUMENTATION_URL="https://docs.voidlinux.org/"
LOGO="void-logo"
ANSI_COLOR="0;38;2;71;128;97"

DISTRIB_ID="void"
```

CarterLi:
My question is that if everyone thinks that the distro should be `Void Linux` instead of `Void`, why did the disto creators set `os-release/NAME` to `Void` instead of `Void Linux`?

For reference, archlinux does set `os-release/Name` to `Arch Linux`

```
NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://bugs.archlinux.org/"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=archlinux-logo
```

CarterLi:
Anyway, thanks for the contribution.

PR #590, opened 2023-10-23
CPU (Linux): quirks for MIPS platforms

matoro:
CPU name is in "cpu model" field on MIPS.  Also, /proc/cpuinfo contains an "isa" field, but rather than being a single string like RISC-V it is instead a space-separated list of every ISA flavor that the chip is compatible with.  Since these are mostly incremental, just make the cpuISA value set to the last one.

Before:
```
$ fastfetch --structure "cpu" --logo none
CPU: mips1 mips2 mips3 mips4 mips5 mips32r1 mips32r2 mips64r1 mips64r2 (16)
```

After:
```
$ fastfetch --structure "cpu" --logo none
CPU: Cavium Octeon II V0.10 mips64r2 (16)
```


CarterLi:
LGTM. Thanks for contribution

PR #591, opened 2023-10-23
Host (Linux): devicetree path for OBP hosts

matoro:
At least SPARC devices have their model name in this devicetree node and do not have a "model" node.  Possibly also POWER Macs, do not have one to test though.


CarterLi:
Thanks

PR #608, opened 2023-11-02
Add another small NixOS logo

ErrorNoInternet:
**It doesn't compile yet**, cmake is changing the codepoints from U+1FB00..U+1FBFF ([Symbols for Legacy Computing](https://en.wikipedia.org/wiki/Symbols_for_Legacy_Computing)) to U+D800..U+DB7F and U+DC00..U+DFFF (surrogate characters?) in `build/logo_builtin.h`

Not sure how to fix this.

New
image

`nixos_small`
image

`nixos`
image

Maybe I should make this `nixos_small` instead of `nixos2_small`? Since it's basically a smaller version of `nixos`.
Also, there seems to be both NixOS_small and NixOsSmall.


CarterLi:
BTW: you may use `--os-keys "os       "` to align your keys

PR #618, opened 2023-11-08
src: end va_list before returning

VoltrexKeyva:
Explicitly end the `va_list` before returning in the `ffStrbufSetF()` function as `va_list`s are not ended implicitly.


CarterLi:
Thanks

PR #624, opened 2023-11-17
Minor cleanups reported by clang-tidy

mame98:
Hey, thanks for creating fastfetch!

I noticed a few issues reported by clang-tidy and attempted to fix some of them. They are mostly from the bugprone- or readability- groups.
Do you think merging those changes is a good idea?

### DIFF
--- a/src/common/io/io_unix.c
+++ b/src/common/io/io_unix.c
@@ -94,7 +94,7 @@ bool ffPathExists(const char* path, FFPathType type)
     if(stat(path, &fileStat) != 0)
         return false;
 
-    int mode = fileStat.st_mode & S_IFMT;
+    unsigned int mode = fileStat.st_mode & S_IFMT;
 
     if(type & FF_PATHTYPE_REGULAR && mode == S_IFREG)
         return true;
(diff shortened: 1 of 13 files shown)


src/detection/vulkan/vulkan.c
```diff
@@ -104,8 +104,8 @@ static const char* detectVulkan(FFVulkanResult* result)
     if(instanceVersion.major == 0 && instanceVersion.minor == 0 && instanceVersion.patch == 0)
         instanceVersion.major = 1;
 
-    VkPhysicalDevice physicalDevices[128];
-    uint32_t physicalDeviceCount = (uint32_t) (sizeof(physicalDevices) / sizeof(*physicalDevices));
+    uint32_t physicalDeviceCount = 128;
+    VkPhysicalDevice physicalDevices[physicalDeviceCount];
```
CarterLi, line 108: Didn't this change make physicalDevices VLA?

mame98: Yeah you are right, I missed that, will push a fix soon

mame98: I simply reverted the original commit

PR #626, opened 2023-11-18
Spellchecking

mame98:
I fixed some typos in the codebase as reported by codespell. I also added a .codespellrc file as well as a workflow that runs codespell.

The only major typo was in [src/modules/gpu/gpu.c](https://github.com/fastfetch-cli/fastfetch/compare/dev...mame98:spellcheck?expand=1#diff-ea1c87371e97d577afe5254a9f69a72aeac60211eb2541b95517e5f02f75fbd5). I think my fix allows both the wrong-spelled config value and the correct one. What do you think?

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -2,4 +2,4 @@
 check-filenames =
 builtin = clear,rare,usage,informal
 skip = */.git,*/cmake-build-*,*/.idea,*/completions,*/presets,*/screenshots,*/tests,*/3rdparty,*/logo/ascii
-ignore-words-list = iterm,compiletime,intergrated,unknwn,pengwin,siduction,master
+ignore-words-list = iterm,compiletime,intergrated,unknwn,pengwin,siduction,master,sur,doas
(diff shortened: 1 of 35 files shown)


.codespellrc
```diff
@@ -0,0 +1,5 @@
+[codespell]
+check-filenames =
+builtin = clear,rare,usage,informal
+skip = */.git,*/cmake-build-*,*/.idea,*/completions,*/presets,*/screenshots,*/tests,*/3rdparty,*/logo/ascii
+ignore-words-list = iterm,compiletime,intergrated,unknwn,pengwin,siduction,master,sur,doas
```
CarterLi, line 5: Fixing `integrated` should be ok. Just remember to mention it in the change log

CarterLi, line 5: Fixing `integrated` should be ok. Just remember to mention it in the change log

mame98: I removed it and added a changelog entry

CarterLi:
Thanks

PR #632, opened 2023-11-27
optimize(IO): speed up createSubfolders

apocelipes:
Avoiding unnecessary ffStrbufAppendC calls to speed up createSubfolders.

Now it finds and adds a subdirectory at a time instead of just appending a char.


apocelipes:
> 1. 运算符两边应该空格
> 2. optimize(io) -> IO

Thanks for reviewing, fixed.

Issue #634, opened 2023-11-27
Fastfetch is slower than Neofetch


flipflop133:
# General description of bug:

* What happened: Fastfetch is slower than Neofetch
* What should happen: Fastfetch should be faster than Neofetch
* Did it work in an older version: first time I use it
* Where did you get the binary: installed `fastfetch 2.2.3-1` with paru
* Does this issue still occurs in [the latest dev build](https://github.com/fastfetch-cli/fastfetch/actions/)? Yes

I think it's slowing down when trying to display the shell informations.
# Comparison image
<img width="1162" alt="image" src="https://github.com/fastfetch-cli/fastfetch/assets/48946818/27130fcc-9785-48d8-a29d-406914f89017">

# Often helpful information:

The content of the configuration file you use (if any)
```
used the default command without any configuration
```

Output of `env NO_CONFIG=1 fastfetch --load-config all --show-errors --stat --multithreading false --disable-linewrap false --hide-cursor false`:

```
$ env NO_CONFIG=1 fastfetch --load-config all --show-errors --stat --multithreading false --disable-linewrap false --hide-cursor false
                  -`                     francois@fox                                                                                             0ms
                 .o+`                    ------------                                                                                             3ms
                `ooo/                    OS: Arch Linux x86_64                                                                                    0ms
               `+oooo:                   Host: 82FG (IdeaPad 5 15ITL05)                                                                           0ms
              `+oooooo:                  Bios (UEFI): FHCN72WW (1.72)                                                                             0ms
              -+oooooo+:                 Board: LNVNB161216 (SDK0J40709 WIN)                                                                      0ms
            `/:-:++oooo+:                Chassis: Notebook (IdeaPad 5 15ITL05)                                                                    0ms
           `/++++/+++++++:               Kernel: 6.6.2-arch1-1                                                                                    0ms
          `/++++++++++++++:              Uptime: 54 mins                                                                                          0ms
         `/+++ooooooooooooo/`            Processes: 296                                                                                           1ms
        ./ooosssso++osssssso+`           Packages: 824 (pacman)                                                                                   1ms
       .oossssso-````/ossssss+`          Shell: bash 5.2.21                                                                                     100ms
      -osssssso.      :ssssssso.         Display (BNQ78F5): 2560x1440 @ 60Hz                                                                    478ms
     :osssssss/        osssso+++.        Brightness (BOE08E2): 0%                                                                                 0ms
[...]

PR #636, opened 2023-11-28
optimize(Displayserver): performance up

apocelipes:
Replace stdio with the posix api, and drop fscanf.

It shows a huge performance up in my local benchmarks.

About 35~40% faster per modefile than the old one.

Here is the highlights of benchmarks:

```text
Running ./bench_linux_amd64
Run on (8 X 1800.01 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x4)
  L1 Instruction 32 KiB (x4)
  L2 Unified 256 KiB (x4)
  L3 Unified 6144 KiB (x1)
Load Average: 0.12, 0.19, 0.16
-------------------------------------------------------
Benchmark             Time             CPU   Iterations
-------------------------------------------------------
BenchStdIO         1712 ns         1712 ns       341076
BenchPOSIXIO        998 ns          998 ns       655856
```

updates #634

### DIFF
--- a/src/detection/displayserver/linux/displayserver_linux.c
+++ b/src/detection/displayserver/linux/displayserver_linux.c
@@ -4,6 +4,10 @@
 #include "util/stringUtils.h"
 
 #include <dirent.h>
+#include <fcntl.h>
+
+// 2 * (uint32_t string max length is 10 bytes) + 'x' + NUL
+#define MODE_FILE_CONTENT_LENGTH 22
 
 static void parseDRM(FFDisplayServerResult* result)
 {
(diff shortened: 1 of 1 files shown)


CarterLi:
Do you really think this PR can address any performance issues?

You are right that parsing string manually is always faster than fscanf, but why fscanf exists?

Come on! 1ms = 1000000ns. Several nanoseconds are not bottleneck, because it's executed only once. In addition, as for #634, this code is NOT executed unless `--ds-force-drm` is used.

Since @flipflop133 was using X11, the real bottleneck was the xrandr stuff. If you really want to improve fastfetch, please check that.

abbycin:
这个pr真是闲的蛋疼，好比是想让sleep

CarterLi:
Don't get me wrong, I like performance improvements, and I'm accepting changes to improve performance at the cost of code complexity, but ONLY in common code.

Closing.

CarterLi:
5x more lines of code and one more magic number are complexity.

> it even costs twice as much in performance

`1712ns -> 998ns` can be called twice.

`500 * 1000 * 1000ns + 1712ns -> 500 * 1000 * 1000ns + 998ns` can't.

Issue #639, opened 2023-11-29
Handling pipe timeout is incorrect


apocelipes:
As argued in #638 , set `O_NONBLOCK` at a pipe which will be dupped to stdout/stderr is unsafe, and has some potential problems. 

Since the code only handle pipes, we can safely use blocking-pipes with `poll` -- There is only one reader, so it couldn't happen that the fd is ready and there is no data to read. (e.g. read by another reader or reported as ready like a socket because of protocol notifications, etc.)

The correct way to do timeout is insteading of using non-blocking io, to use a big enough buff (as big as the buffer inside the pipe) to read data from pipes, and call `poll` again no matter how much data is read. When `read` returns EOF or error, break out the loop. A big enough buffer is to reduce `read` calls, a 4096 length buffer works too but may be encounter some performance impacts.

`O_NONBLOCK` problems should be fixed.

Additionally pipe needs to set the O_CLOEXEC flag for the most important reason, besides performance and preventing file descriptor leaks, for security. We have no control over the external programs that are called, so it is important to minimize the amount of information that is exposed.

CarterLi:
> Additionally pipe needs to set the O_CLOEXEC flag for the most important reason, besides performance and preventing file descriptor leaks, for security.

[They were closed](https://github.com/fastfetch-cli/fastfetch/blob/dev/src/common/processing_linux.c#L40-L41). However, you are right, `O_CLOEXEC` is a better option. In addition, we should check every `open(2)` syscall to have `O_CLOEXEC` used.

> `O_NONBLOCK` problems should be fixed.

[If timeout is not enabled, `O_NONBLOCK` won't be used.](https://github.com/fastfetch-cli/fastfetch/blob/dev/src/common/processing_linux.c#L52-L54) I don't see any problems here.

apocelipes:
> [If timeout is not enabled, `O_NONBLOCK` won't be used.](https://github.com/fastfetch-cli/fastfetch/blob/dev/src/common/processing_linux.c#L52-L54) I don't see any problems here.

If timeout set, fcntl will also change child process's pipe flags, this can be privode by this:

```c
#include <fcntl.h>
#include <unistd.h>
#include <stdio.h>

int main()
{
        int fds[2] = {-1, -1};
        pipe(fds);
        int pid = fork();
        if (pid) {
                usleep(10000000);
                printf("before: %d\n", fcntl(fds[1], F_GETFL));
[...]

PR #640, opened 2023-11-29
fix(Common): fixes for ffProcessAppendOutput

apocelipes:
- Use pipe2 on Linux and BSDs, which can set all flags through a single syscall.
- macOS/darwin does not support pipe2, add a wrapper for it.
- Close pipes when fork() failed. These fds should be closed when the function returns.
- Set O_CLOEXEC at pipes, the exec() family of functions will close this fds automatically.
- Use _exit instead of exit. Exit calls the callbacks registered by atexit and flushes the stdio's buffer, which can lead to some unexpected behavior, especially after an exec syscall failure.
- Fix timeout.
- Not use non-blocking IO. Because we only wait for one file, and cannot do other things during waiting.

For #639

### DIFF
--- a/src/common/processing_linux.c
+++ b/src/common/processing_linux.c
@@ -11,7 +11,7 @@
 #include <errno.h>
 #include <sys/wait.h>
 
-enum { FF_PIPE_BUFSIZ = 4096 };
+#define FF_PIPE_BUFSIZ 4096
 
 static inline void waitpid_wrapper(const pid_t* pid)
 {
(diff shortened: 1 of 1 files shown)


src/common/processing_linux.c
```diff
@@ -11,7 +11,7 @@
 #include <errno.h>
 #include <sys/wait.h>
 
-enum { FF_PIPE_BUFSIZ = 4096 };
+#define FF_PIPE_BUFSIZ 16384
```
CarterLi, line 14: You once again assume unreliable things. The pipe buf size has been increased twice, how can you be sure that it won't be increased again?

However, I think it should be ok even if FF_PIPE_BUFSIZ is smaller than real pipe buf size.

src/common/processing_linux.c
```diff
@@ -20,40 +20,52 @@ static inline void waitpid_wrapper(const pid_t* pid)
         waitpid(*pid, NULL, 0);
 }
 
+static inline int ffPipe2(int *fds, int flags) {
```
CarterLi, line 23: Please follow the coding style of fastfetch. Move `{` to a new line

src/common/processing_linux.c
```diff
@@ -20,40 +20,52 @@ static inline void waitpid_wrapper(const pid_t* pid)
...
+    #else
+        return pipe2(fds, flags);
+    #endif
+}
+
 const char* ffProcessAppendOutput(FFstrbuf* buffer, char* const argv[], bool useStdErr)
 {
     int pipes[2];
+    const int timeout = instance.config.general.processingTimeout;
+    const int pipeFlags = O_CLOEXEC;
```
CarterLi, line 39: Don't make it a separate variable. It's used only once and `const` in `C` doesn't really make a variable a constant.

src/common/processing_linux.c
```diff
@@ -20,40 +20,52 @@ static inline void waitpid_wrapper(const pid_t* pid)
...
 
-    if(pipe(pipes) == -1)
+    if(ffPipe2(pipes, pipeFlags) == -1)
         return "pipe() failed";
 
     __attribute__((__cleanup__(waitpid_wrapper))) pid_t childPid = fork();
-    if(childPid == -1)
+    if(childPid == -1) {
```
CarterLi, line 45: ditto

src/common/processing_linux.c
```diff
@@ -20,40 +20,52 @@ static inline void waitpid_wrapper(const pid_t* pid)
...
+        close(pipes[0]);
+        close(pipes[1]);
         return "fork() failed";
+    }
 
     //Child
     if(childPid == 0)
     {
         int nullFile = open("/dev/null", O_WRONLY);
```
CarterLi, line 54: This is where the bug is. It should be flagged as O_CLOEXEC

src/common/processing_linux.c
apocelipes: > You once again assume unreliable things. The pipe buf size has been increased twice, how can you be sure that it won't be increased again?
> 
> However, I think it should be ok even if FF_PIPE_BUFSIZ is smaller than real pipe buf size.

It's not my assume, 16384 is the default size of macOS/BSDs pipe buffer.

If this confuse you, i'll change it back to 4096.

src/common/processing_linux.c
apocelipes: Fixed.

src/common/processing_linux.c
apocelipes: Fixed.

src/common/processing_linux.c
apocelipes: Fixed.

src/common/processing_linux.c
apocelipes: Fixed.

PR #642, opened 2023-12-03
optimize(IO): reduce memory usage during reading data from files to F…

apocelipes:
…Fstrbufs

Avoid unnecessary calls of ffStrbufEnsureFree and only allocate necessary memory.

In most cases, this will save 30% ~ 50% of memory used by ffAppendFDBuffer.


src/common/io/io_windows.c
```diff
@@ -37,17 +37,26 @@ bool ffAppendFDBuffer(HANDLE handle, FFstrbuf* buffer)
...
+    if (fileSize.QuadPart > 0)
+    {
+        // optimize for files has a fixed length,
+        // file can be very large, only keep necessary memory to save time and resources.
+        ffStrbufEnsureFixedLengthFree(buffer, (uint32_t)fileSize.QuadPart);
+    }
+    else
+        ffStrbufEnsureFree(buffer, 31);
     uint32_t free = ffStrbufGetFree(buffer);
+    int64_t remain = fileSize.QuadPart; // Windows does not have ssize_t, use int64_t to instead.
```
CarterLi, line 50: > Windows does not have ssize_t, use int64_t to instead

Source?

src/util/FFstrbuf.c
```diff
@@ -54,6 +54,30 @@ void ffStrbufEnsureFree(FFstrbuf* strbuf, uint32_t free)
...
+    uint32_t newCap = strbuf->allocated + (free - oldFree);
+    
+    if(strbuf->allocated == 0)
+    {
+        ++newCap; // for the NUL
+        char* newbuf = malloc(sizeof(*strbuf->chars) * newCap);
+        if(strbuf->length == 0)
+            *newbuf = '\0';
+        else
+            memcpy(newbuf, strbuf->chars, strbuf->length + 1);
```
CarterLi, line 72: What if `strbuf->length > free`?

src/common/io/io_windows.c
apocelipes: https://learn.microsoft.com/en-us/windows/win32/winprog/windows-data-types#ssize_t

Windows has SSIZE_T and the defination is actually difference from the *nix.

I'll try this: https://github.com/ycm-core/YouCompleteMe/issues/3949#issuecomment-944786867

src/util/FFstrbuf.c
apocelipes: length should be checked at the first:

```c
if ((oldFree >= free && !(strbuf->allocated == 0 && strbuf->length > 0)) || strbuf->length >= free)
        return;
```

src/common/io/io_windows.c
CarterLi: Did you search ssize_t in fastfetch code base?

apocelipes: Oh, there is already a ssize_t in io_windows.c, i'm sorry :P

src/util/FFstrbuf.c
```diff
@@ -54,6 +54,30 @@ void ffStrbufEnsureFree(FFstrbuf* strbuf, uint32_t free)
     strbuf->allocated = allocate;
 }
 
+void ffStrbufEnsureFixedLengthFree(FFstrbuf* strbuf, uint32_t free)
```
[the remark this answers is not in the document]
apocelipes: Testcases have been added. Fixed.

PR #644, opened 2023-12-04
Fix procfs reading

apocelipes:
Files in procfs are always changed by kernel. For example: The contents of `/proc/meminfo` and `/proc/stat`  are changed frequently to show the "real-time" status in the system.

It means if we read the data with more than one read syscall, we could get an incorrect data which was broken by kernel.

A typical case about this problem was reported by psutil in https://github.com/giampaolo/psutil/issues/2050

Using an big buffer let read reads the whole file can fix this.

psutil uses a 32K buffer while procps-ng uses a 8K buffer, I think a 8K buffer is enough.

Since we read the whole file's data, parsing can be simpler than before.

Using of stack buffers and getting rid of stdio also saves about 30 heap allocations.

Files like `/proc/cpuinfo`, `/proc/mounts` whose contents don't get changed very often don't need to be fixed, although psutils best practices tell us we should read all the data we need in one read syscall.

### DIFF
--- a/src/detection/cpuusage/cpuusage_linux.c
+++ b/src/detection/cpuusage/cpuusage_linux.c
@@ -41,7 +41,7 @@ const char* ffGetCpuUsageInfo(FFlist* cpuTimes)
             };
         }
         else
-            break; // because we read the whole /proc/stat, we can safely quit when the line does not start with "cpuN" 
+            break;
         start = token + 1;
     }
 
--- a/src/detection/cpuusage/cpuusage_linux.c
+++ b/src/detection/cpuusage/cpuusage_linux.c
@@ -7,30 +7,42 @@
 
 const char* ffGetCpuUsageInfo(FFlist* cpuTimes)
 {
-    FF_AUTO_CLOSE_FILE FILE* procStat = fopen("/proc/stat", "r");
-    if(procStat == NULL)
+    char buf[PROC_FILE_BUFFSIZ];
+    ssize_t nRead = ffReadFileData("/proc/stat", sizeof(buf) - 1, buf);
+    if(nRead < 0)
     {
         #ifdef __ANDROID__
         return "Accessing \"/proc/stat\" is restricted on Android O+";
         #else
-        return "fopen(\"""/proc/stat\", \"r\") == NULL";
+        return "ffReadFileData(\"/proc/stat\", sizeof(buf) - 1, buf) failed";
         #endif
     }
+    buf[nRead] = '\0';
+
     // Skip first line
-    if (fscanf(procStat, "cpu%*[^\n]\n") < 0)
-        return "fscanf() first line failed";
+    char *start = NULL;
+    if((start = strchr(buf, '\n')) == NULL)
+        return "skip first line failed";
+    ++start;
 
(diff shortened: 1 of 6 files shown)


src/detection/cpuusage/cpuusage_linux.c
```diff
@@ -7,30 +7,42 @@
...
-            .totalAll = (uint64_t)total,
-        };
+            FFCpuUsageInfo* info = (FFCpuUsageInfo*) ffListAdd(cpuTimes);
+            *info = (FFCpuUsageInfo) {
+                .inUseAll = (uint64_t)inUse,
+                .totalAll = (uint64_t)total,
+            };
+        }
+        else
+            break; // because we read the whole /proc/stat, we can safely quit when the line does not start with "cpuN"
```
CarterLi, line 44: Does it really matter whether we read the whole file or line by line?

src/detection/swap/swap_linux.c
```diff
@@ -8,25 +8,21 @@
...
+    buf[nRead] = '\0';
 
     uint64_t swapTotal = 0, swapFree = 0;
 
-    char* FF_AUTO_FREE line = NULL;
-    size_t len = 0;
-    uint8_t count = 0;
+    char *token = NULL;
+    if ((token = strstr(buf, "SwapTotal:")) != NULL)
+        sscanf(token, "SwapTotal: %" PRIu64, &swapTotal);
```
CarterLi, line 21: ditto

src/detection/uptime/uptime_linux.c
```diff
@@ -7,14 +7,17 @@
...
-    if (!uptime) return "fopen(\"/proc/uptime\", \"r\") failed";
+    char buf[64];
+    ssize_t nRead = ffReadFileData("/proc/uptime", sizeof(buf) - 1, buf);
+    if(nRead < 0)
+        return "ffReadFileData(\"/proc/uptime\", sizeof(buf) - 1, buf) failed";
+    buf[nRead] = '\0';
 
     double sec;
-    if (fscanf(uptime, "%lf", &sec) > 0)
+    if(sscanf(buf, "%lf", &sec) > 0)
```
CarterLi, line 17: strtod

src/detection/swap/swap_linux.c
apocelipes: Fixed.

src/detection/memory/memory_linux.c
```diff
@@ -6,45 +6,38 @@
...
              shmem = 0,
              memFree = 0,
              buffers = 0,
              cached = 0,
              sReclaimable = 0;
-    uint8_t count = 0;
+    
+    char *token = NULL;
+    if((token = strstr(buf, "MemTotal:")) != NULL)
+        sscanf(token, "MemTotal: %" PRIu64, &memTotal);
```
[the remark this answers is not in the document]
apocelipes: Fixed.

src/detection/cpuusage/cpuusage_linux.c
apocelipes: Removed. Fixed.

PR #652, opened 2023-12-07
optimize(Packages): use mmap to read large files

apocelipes:
Using less syscalls and processing data in memory makes getNumStringsImpl ~400% faster.

Some benchmark highlights:

On my Ubuntu 22.04 laptop, with 1408 dpkg packages.

```text
Run on (8 X 2400.01 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x4)
  L1 Instruction 32 KiB (x4)
  L2 Unified 256 KiB (x4)
  L3 Unified 8192 KiB (x1)
Load Average: 0.12, 0.06, 0.01
---------------------------------------------------------------------
Benchmark                           Time             CPU   Iterations
---------------------------------------------------------------------
BM_getNumStringsImpl           256691 ns       256694 ns         2517
BM_getNumStringsLegacyImpl    1080674 ns      1080676 ns          636
```

### DIFF
--- a/src/detection/packages/packages_linux.c
+++ b/src/detection/packages/packages_linux.c
@@ -42,25 +42,18 @@ static uint32_t getNumElements(FFstrbuf* baseDir, const char* dirname, unsigned 
 
 static uint32_t getNumStringsImpl(const char* filename, const char* needle)
 {
-    FILE* file = fopen(filename, "r");
-    if(file == NULL)
+    FF_STRBUF_AUTO_DESTROY content = ffStrbufCreate();
+    if (!ffReadFileBuffer(filename, &content))
         return 0;
 
     uint32_t count = 0;
-
-    char* line = NULL;
-    size_t len = 0;
-
-    while(getline(&line, &len, file) != EOF)
+    char *iter = content.chars;
+    size_t needleLength = strlen(needle);
+    while ((iter = memmem(iter, content.length - (size_t)(iter - content.chars), needle, needleLength)) != NULL)
     {
-        if(strstr(line, needle) != NULL)
-            ++count;
+        ++count;
+        iter += needleLength;
     }
-
-    if(line != NULL)
-        free(line);
-
-    fclose(file);
 
     return count;
 }


src/detection/packages/packages_linux.c
```diff
@@ -65,6 +70,44 @@ static uint32_t getNumStringsImpl(const char* filename, const char* needle)
...
+    struct stat fileStat;
+    if(fstat(fd, &fileStat) < 0)
+        return 0;
+
+    size_t fileSize = (size_t)fileStat.st_size;
+    if(fileStat.st_size <= 0 || SIZE_MAX - 4095 <= fileSize)
+    {
+        // file size is zero or too big for mmap
+        // falling back to the line by line implementation
+        return getNumStringsLegacyImpl(filename, needle);
```
CarterLi, line 88: use fdopen?

src/detection/packages/packages_linux.c
```diff
@@ -65,6 +70,44 @@ static uint32_t getNumStringsImpl(const char* filename, const char* needle)
...
 
+static uint32_t getNumStringsImpl(const char* filename, const char* needle)
+{
+    int fd = open(filename, O_RDONLY | O_CLOEXEC);
+    if(fd < 0)
+        return 0;
+
+    struct stat fileStat;
+    if(fstat(fd, &fileStat) < 0)
+        return 0;
```
CarterLi, line 81: leak

src/detection/packages/packages_linux.c
apocelipes: fdopen will take the fd's ownership, I think there should be no problem in adding comments to explain it.
Fixed.

src/detection/packages/packages_linux.c
```diff
@@ -65,6 +70,44 @@ static uint32_t getNumStringsImpl(const char* filename, const char* needle)
...
+    {
+        // file size is zero or too big for mmap
+        // falling back to the line by line implementation
+        return getNumStringsLegacyImpl(filename, needle);
+    }
+
+    const size_t mapSize = (fileSize + 4095) & ( ~ (size_t)4095);
+    char *buf = mmap(NULL, mapSize, PROT_READ, MAP_PRIVATE, fd, 0);
+    if (buf == MAP_FAILED)
+        return 0;
```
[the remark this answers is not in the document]
apocelipes: Fixed.

src/detection/packages/packages_linux.c
apocelipes: Fixed.

CarterLi:
Can't we use FF_AUTO_CLOSE_FD?

CarterLi:
Also, all these early returns should fall back to the old function

apocelipes:
> Also, all these early returns should fall back to the old function

I don't think fdopen can work if fstat already failed.

mmap is a bit complicated, and there are a few errors where it can fall back. This kind of situation is quite rare. The only thing that is more likely to happen is mmap failure due to no enough memory. Is it really worth falling back here?

CarterLi:
I am actually not familiar with mmap.

Did you test ffReadFileBuffer? How does ffReadFileBuffer compare to mmap?

apocelipes:
> I am actually not familiar with mmap.
> 
> Did you test ffReadFileBuffer? How does ffReadFileBuffer compare to mmap?

I'll add a benchmark with ffReadFileBuffer.

CarterLi:
Also, you should use benchmark fastfetch itself, not pull things out. I suggest hyperfine. If the benchmark result differs not much, I prefer use the simpler implementation.

apocelipes:
> I am actually not familiar with mmap.
> 
> Did you test ffReadFileBuffer? How does ffReadFileBuffer compare to mmap?

BM_getNumStringsImpl                    1400988 ns      1400979 ns          476
BM_getNumStringsLegacyImpl              6468152 ns      6468036 ns          108
BM_getNumStringsWithffReadFileBuffer    1839101 ns      1839052 ns          376

Seems like 30% slower than mmap.

the code:

```c
static uint32_t useFFReadBuffer(const char* filename, const char* needle)
{
    FF_STRBUF_AUTO_DESTROY content = ffStrbufCreate();
    if (!ffReadFileBuffer(filename, &content))
        return 0;
    uint32_t count = 0;
    size_t needleLength = strlen(needle);
    char *iter = content.chars;
    while ((iter = strstr(iter, needle)) != NULL)
    {
	++count;
        iter += needleLength;
    }
    return count;
}
```

CarterLi:
What about ReadFileBuffer?

apocelipes:
> What about ReadFileBuffer?

single module: 
image

image

entire command:

image

Not as much fast as mmap, but is still faster than stdio.

CarterLi:
You can use `hyperfine './ff-stdio -s package' './ff-buffer -s package' './ff-mmap -s package'`

I prefer `ff-buffer` since its implementation is the simplest.

apocelipes:
> You can use `hyperfine './ff-stdio -s package' './ff-buffer -s package' './ff-mmap -s package'`
> 
> I prefer `ff-buffer` since its implementation is the simplest.

image

Faster than mmap may be an inaccuracy.

By the way, my benchmarking environment only has 1410 packages, whether other distros could have 2000+ packages. Theoretically the more packages you have the more significant the performance difference.

If u prefer ffReadBuffer, I'll change to it later, and getNumStringsLegacyImpl can be deleted.

CarterLi:
> If u prefer ffReadBuffer, I'll change to it later, and getNumStringsLegacyImpl can be deleted

Please do it.

And you may use `memmem` instead of `strstr`, since lengths of both string are known.

apocelipes:
> Please do it.

Done. Please take a look.

PR #659, opened 2023-12-14
remove unnecessary escaping for Adelie logo

mc680x0:
Sadly, a logo comprised largely of characters that need to be escaped looks really weird on `flashfetch`. 

I was responsible for the PR to hyfetch that resulted in Adelie being added, so it seemed I should clean up on fastfetch also.

I trust this will prove useful.


CarterLi:
Thanks

PR #663, opened 2023-12-18
Minor fixes

apocelipes:
- fix ffPathExists: only using stat makes FF_PATHTYPE_LINK never work under *nix systems. Some symlinks are linked to directories, so keep use stat when FF_PATHTYPE_LINK is not contained in `type` paramater for backward compatibility.
- ffWriteFileData only retries when error is ENOENT. Other errors such as EACCESS, ELOOP or EDQUOT also fail mkdir(), retrying after these errors is useless.
- fix readlink usage: readlink won't append a NUL, we should check the return value and add a NUL handly.

### DIFF
--- a/src/common/io/io.h
+++ b/src/common/io/io.h
@@ -10,6 +10,7 @@
 #else
     #include <unistd.h>
     #include <dirent.h>
+    #include <errno.h>
     typedef int FFNativeFD;
     // procfs's file can be changed between read calls such as /proc/meminfo and /proc/uptime.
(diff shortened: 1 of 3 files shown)


apocelipes:
Would you tell why close this?

I think readlink's call is actually a bug.

CarterLi:
See https://github.com/fastfetch-cli/fastfetch/commit/4ae0c43c6f72173719322ae9e9ac093417a583ef

PR #664, opened 2023-12-18
fix(DiskIO): fix readlink

apocelipes:


CarterLi:
You fix what?

apocelipes:
`char pathSysDeviceReal[PATH_MAX] = "";`

equals to

`char pathSysDeviceReal[PATH_MAX] = { '\0' };`

and the manpages said: 
> readlink() does not append a null byte to buf.

So, if readlink success, we got a invalid Null-terminated string.

apocelipes:
Or, if use this:

`char pathSysDeviceReal[PATH_MAX] = { '\0' };`

It will zero-filled by memset or other functions. Since PATH_MAX on some Linux distros are set to 4096, it could be a noticeable cost.

CarterLi:
image

In <https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1124.pdf>, Page 127

apocelipes:
> image
> 
> In https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1124.pdf, Page 127

This since C11, this project using C11, so, it will be zero-filled, and in most cases it's unnecessary.

CarterLi:
> This since C11

Citation needed

CarterLi:
And even if you want to "fix" this, the same code in physical_linux.c (committed today) should be fixed too.

CarterLi:
Please be cautious when using words like "bug" and "fix"

apocelipes:
> > This since C11
> 
> Citation needed

It's my miss, C99 said the same thing at 6.7.8-21, some old unix's CCs confuse me with their implement behaviors.

apocelipes:
> Please be cautious when using words like "bug" and "fix"

Thank u, I see.

PR #667, opened 2023-12-19
refactor(WMTheme): clean up detectOpenbox

apocelipes:
Simplify the code and implement stricter checking of xml contents.


CarterLi:
Seems good.

PR #668, opened 2023-12-20
Support for looking up user-specific configuration files from `platform.configDirs`

apocelipes:
And implement TODOs in TerminalFont and WMTheme modules.


CarterLi:
Thanks

PR #671, opened 2023-12-22
refactor(IO): clean up ffAppendFDBuffer

apocelipes:
for 2 reasons:

- Simplify the Code.
- ffAppendFDBuffer can't make assumptions about how the read data will be processed, so the trim calls are inappropriate. All callers currently handle whitespace and newlines at the end of files themselves. So these trims are also redundant and can simply be removed.

### DIFF
--- a/src/common/io/io_unix.c
+++ b/src/common/io/io_unix.c
@@ -47,45 +47,45 @@ bool ffWriteFileData(const char* fileName, size_t dataSize, const void* data)
     return write(fd, data, dataSize) > 0;
 }
 
+static inline void readWithLength(int fd, FFstrbuf* buffer, uint32_t length)
+{
+    ffStrbufEnsureFixedLengthFree(buffer, length);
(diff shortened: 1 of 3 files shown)


apocelipes:
Comment out that ffStrbufEnsureFixedLengthFree will take care of NUL.

CarterLi:
I think you are correct. Thanks

CarterLi:
image
image

Maybe need some fixes @apocelipes

CarterLi:
It turns out that there are so many trailing line endings in sysfs. I fixed some of them, but surely not all. Although the change is right, it can be too dangerous.

apocelipes:
Yes. There are 40+ calls need to be checked. Maybe we should bring trims back and add a todo: "remove this after all check finished".

apocelipes:
@CarterLi I need to apologize for not checking the code of some modules carefully. Now I did a full code review and created a new PR: #673 . Please take a look.

PR #679, opened 2024-01-01
Add an option to include building the extra NVIDIA GPU support

ceamac:
I noticed `nvml.h` is not open source, so I'd like to be able to exclude it from the build if needed.

(BTW, Intel is in the same situation with `igcl_api.h`, but that is used only on Windows.)


CarterLi:
> I noticed `nvml.h` is not open source

I'm not a license expert. NVIDIA developers themselves embeded `nvml.h` in at least an [Apache-2.0 project](https://github.com/NVIDIA/DCGM/blob/master/sdk/nvidia/nvml/nvml.h) and [GPL-2.0 project](https://github.com/NVIDIA/nvidia-settings/blob/main/src/nvml.h) which means that the license of `nvml.h` should be looser than Apache-2.0 and GPL-2.0. In addition, the `nvml.h` file is provided as source officially in the [NVIDIA SDK](https://developer.nvidia.com/tesla-deployment-kit). So that I think the license embeded in `nvml.h` source is loose enough to be used in other open source projects.

In addition, `nvml.h` contains only function declarations, which doesn't actually generate code. So that the binaries compiled with `nvml.h` doesn't actually contains code of `nvml`.

I'm aware of the license concerns. The NVML lib binaries are the part of the proprietary NVIDIA graphic driver which are closed source, but the header file `nvml.h` is not, in my opinion.

CarterLi:
Another option is writing it myself by referring to the [document](https://docs.nvidia.com/deploy/nvml-api/index.html). The result should not make much difference with the official one.

ceamac:
I'm not a license expert either, I just read the copyright notice at the top of `nvml.h` and it got me concerned.  I think "nonexclusive, royalty-free license to use this code in individual and commercial software" is good enough to be able to use it, but it still is not a free software license (like approved by FSF or OSI).

Anyway, we have this snipped in the [Gentoo devmanual](https://devmanual.gentoo.org/general-concepts/licenses/#determining-the-correct-license) that basically says the individual license notice in each source file trumps the package's general license.  Nvidia may include this file anywhere they want, as they own the copyright, but the rest of the world must respect their license.

With this in mind, I want to add this Nvidia license to the package and let open source purists opt out of using it.  But I can understand if you don't want to accept this patch, it's quite niche and debatable.

Thank you anyway.

CarterLi:
> Nvidia may include this file anywhere they want, as they own the copyright

I don't think so. People who created the law should also obey the law. Nvidia puts the file in GPL-2.0 and Apache-2.0 projects means that they admit that the license of `nvml.h` is GPL-2.0 and Apache-2.0 compatible.

They even release its [GO bindings in Apache-2.0](https://github.com/NVIDIA/go-nvml). There is no way that the license of original project is tighter than the one of derived project.

Anyway, I'm aware of the license concerns of package managers. One should have the way to drop things don't want. I just want to argue that the license should not be a problem.

PR #682, opened 2024-01-12
A little clean up

apocelipes:
Changing recursion into a loop.

### DIFF
--- a/src/detection/terminalshell/terminalshell_linux.c
+++ b/src/detection/terminalshell/terminalshell_linux.c
@@ -125,33 +125,37 @@ static pid_t getShellInfo(FFShellResult* result, pid_t pid)
 
     pid_t ppid = 0;
 
-    if(getProcessNameAndPpid(pid, name, &ppid))
-        return 0;
+    while (getProcessNameAndPpid(pid, name, &ppid) == NULL)
+    {
+        //Common programs that are between terminal and own process, but are not the shell
+        if(
+            ffStrEquals(name, "sh")                  || //This prevents us from detecting things like pipes and redirects, i hope nobody uses plain `sh` as shell
+            ffStrEquals(name, "sudo")                ||
+            ffStrEquals(name, "su")                  ||
+            ffStrEquals(name, "strace")              ||
+            ffStrEquals(name, "sshd")                ||
+            ffStrEquals(name, "gdb")                 ||
+            ffStrEquals(name, "lldb")                ||
+            ffStrEquals(name, "lldb-mi")             ||
+            ffStrEquals(name, "login")               ||
+            ffStrEquals(name, "ltrace")              ||
+            ffStrEquals(name, "perf")                ||
+            ffStrEquals(name, "guake-wrapped")       ||
+            ffStrContainsIgnCase(name, "debug")      ||
+            ffStrContainsIgnCase(name, "not-found")  ||
+            ffStrEndsWith(name, ".sh")
+        )
+        {
+            pid = ppid;
+            continue;
+        }
 
-    //Common programs that are between terminal and own process, but are not the shell
-    if(
-        ffStrEquals(name, "sh")                  || //This prevents us from detecting things like pipes and redirects, i hope nobody uses plain `sh` as shell
-        ffStrEquals(name, "sudo")                ||
-        ffStrEquals(name, "su")                  ||
-        ffStrEquals(name, "strace")              ||
-        ffStrEquals(name, "sshd")                ||
-        ffStrEquals(name, "gdb")                 ||
-        ffStrEquals(name, "lldb")                ||
-        ffStrEquals(name, "lldb-mi")             ||
-        ffStrEquals(name, "login")               ||
-        ffStrEquals(name, "ltrace")              ||
-        ffStrEquals(name, "perf")                ||
-        ffStrEquals(name, "guake-wrapped")       ||
-        ffStrContainsIgnCase(name, "debug")      ||
-        ffStrContainsIgnCase(name, "not-found")  ||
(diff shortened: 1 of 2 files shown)


CarterLi:
Why bothering? It's tail recursion and will be compiled into a loop.

apocelipes:
The old code has 4 return points, the loop has only one return point, which is logically easier to understand.

And I checked the asm generated by gcc (11.4 Ubuntu), most of the recursion were optimized to a `jmp`, but still a `callq getTerminalShell` left. Perhaps using -O3 would further optimize away this call, but converting to an equivalent loop might be more appropriate.

apocelipes:
The old one's asm looks like this:

```asm
000000000009b500 <getTerminalShell.constprop.0>:
   ...
   9b853:	4c 89 ef             	mov    %r13,%rdi
   9b856:	e8 15 04 f8 ff       	call   1bc70 <ffStrbufAppendS.part.0.lto_priv.0.lto_priv.0>
   9b85b:	48 8d 0d 9e 6b 07 00 	lea    0x76b9e(%rip),%rcx        # 112400 <result.0.lto_priv.3+0x20>
   9b862:	4c 89 ee             	mov    %r13,%rsi
   9b865:	89 ef                	mov    %ebp,%edi
   9b867:	48 8d 51 f0          	lea    -0x10(%rcx),%rdx
   9b86b:	e8 b0 cf fd ff       	call   78820 <getProcessInformation.lto_priv.0>
   9b870:	8b 7c 24 0c          	mov    0xc(%rsp),%edi
   9b874:	e8 87 fc ff ff       	call   9b500 <getTerminalShell.constprop.0>
   9b879:	e9 cb fc ff ff       	jmp    9b549 <getTerminalShell.constprop.0+0x49>
   9b87e:	e8 ad 75 f7 ff       	call   12e30 <__stack_chk_fail@plt>
   9b883:	4c 8d 2d a6 6b 07 00 	lea    0x76ba6(%rip),%r13        # 112430 <result.0.lto_priv.3+0x50>
   9b88a:	89 2d e8 6b 07 00    	mov    %ebp,0x76be8(%rip)        # 112478 <result.0.lto_priv.3+0x98>
   9b890:	4c 89 ef             	mov    %r13,%rdi
   ...
```

The C code all looks like normal tail recursion, but the compiler may have its own decisions to make.

CarterLi:
Code was updated. You may rebase it and update the windows one too.

apocelipes:
Updated. Please take a look.

CarterLi:
Well I still don't like the change. The code is not simplified since you basically change `return` to `continue` and `break`. In addition, you have to maintain the state variables manually like `pid = ppid` which make the code a little harder to understand IMO.

src/detection/terminalshell/terminalshell_linux.c
```diff
@@ -161,39 +166,47 @@ static pid_t getTerminalInfo(FFTerminalShellResult* result, pid_t pid)
...
-    // https://github.com/fastfetch-cli/fastfetch/discussions/501
-    if (ffStrEndsWith(name, " (figterm)") || ffStrEndsWith(name, " (cwterm)"))
-        getProcessNameAndPpid(ppid, name, &ppid);
-    #endif
+        #ifdef __APPLE__
+        // https://github.com/fastfetch-cli/fastfetch/discussions/501
+        if (ffStrEndsWith(name, " (figterm)") || ffStrEndsWith(name, " (cwterm)"))
+        {
+            if (getProcessNameAndPpid(ppid, name, &ppid) != NULL)
+                return 0;
```
CarterLi, line 200: Do you really understand what the code does?

apocelipes:
> Well I still don't like the change. The code is not simplified since you basically change `return` to `continue` and `break`. In addition, you have to maintain the state variables manually like `pid = ppid` which make the code a little harder to understand IMO.

In the recursion we also trace the state variables manually as the paramater, and we also need to backtrack function executions because of 4 different exit points, this makes the code logic not linear.

The indentation due to loops may make you feel a little weird, but the circle complexity doesn't go up as a result.

src/detection/terminalshell/terminalshell_linux.c
CarterLi: > `ffReadFileData` would failed by no memory or other IO errors

In macOS we don't use `ffReadFileData`. But okey, this check makes at least no harm.

CarterLi:
> because of 4 different exit points, this makes the code logic not linear.

Changing `return` to `break` or `continue` doesn't make che code logic linear. [Return early](https://medium.com/swlh/return-early-pattern-3d18a41bba8) is a good programing practice to reduce the code nesting.

apocelipes:
> Changing `return` to `break` or `continue` doesn't make che code logic linear. [Return early](https://medium.com/swlh/return-early-pattern-3d18a41bba8) is a good programing practice to reduce the code nesting.

But your "return early" leds to the recursion, I don't think this is good practice for this idiom. continue and break limit the logic to while loops, which is a little simpler than between function calls. The latter point is my personal opinion.

CarterLi:
> In the recursion we also trace the state variables manually as the paramater

Right, but calling a function makes a new context.

For example

```
// use loop
void fn() {
    int state = 0;
    while (something)
    {
        ++state;
        // use state
    }
    // oops, state has been changed
}
```

```
// use recursion call
void fn(int state) {
    if (something)
    {
        return fn(state + 1)
    }
    // ok, state is preserved
    // do something with state
}
```

CarterLi:
Okey. I will merge this if you resolve the conflicts.

apocelipes:
The conflicts have been resolved. Please take a look.

PR #706, opened 2024-01-30
CPU (Linux): use marketing name on Asahi Linux

ydalton:
I saw your reddit post on the Asahi Linux subreddit and wanted to help out. I admit it's not the best code but I don't think there is another way. Since you have an M1 Pro, can you check if it reports the correct CPU model?

### DIFF
--- a/src/detection/cpu/cpu_linux.c
+++ b/src/detection/cpu/cpu_linux.c
@@ -197,6 +197,87 @@ const char* ffDetectCPUImpl(const FFCPUOptions* options, FFCPUResult* cpu)
                 }
             }
 
+            #ifdef __aarch64__
+            // In Asahi Linux, reading /proc/device-tree/compatible gives
+            // information on the device model. It consists of 3 NUL terminated
+            // strings, the second of which gives the actual SoC model. But it
+            // is not the marketing name, i.e. for M2 there is "apple,t8112" in
+            // the compatible string.
+            //
+            // A full list of the SoC model names can be found here:
+            // https://github.com/AsahiLinux/docs/wiki/Codenames
+            #define DT_COMPAT_PATH "/proc/device-tree/compatible"
+            if (ffStrbufEqualS(&cpu->vendor, "Apple"))
+            {
+                FF_STRBUF_AUTO_DESTROY content = ffStrbufCreate();
+                char* cpu_compat = NULL;
+                char* cpu_model_name = NULL;
+
+                ffStrbufAppend(&cpu->name, &cpu->vendor);
+                ffStrbufAppendC(&cpu->name, ' ');
+
+                if (ffAppendFileBuffer(DT_COMPAT_PATH, &content))
+                {
+                    // get the second NUL terminated string
+                    cpu_compat = content.chars + strlen(content.chars) + 1;
+                    // "apple,t8112" -> "t8112"
+                    cpu_compat = strchr(cpu_compat, ',') + 1;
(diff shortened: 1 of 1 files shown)


src/detection/cpu/cpu_linux.c
```diff
@@ -197,6 +197,86 @@ const char* ffDetectCPUImpl(const FFCPUOptions* options, FFCPUResult* cpu)
...
+                        cpu_model_name = "M3";
+                    }
+                    else if (strcmp(cpu_compat, "t6030") == 0)
+                    {
+                        cpu_model_name = "M3 Pro";
+                    }
+                    else if (strcmp(cpu_compat, "t6031") == 0
+                                    || strcmp(cpu_compat, "t3034") == 0)
+                    {
+                        cpu_model_name = "M3 Pro";
```
CarterLi, line 267: M3 Max?

src/detection/cpu/cpu_linux.c
```diff
@@ -197,6 +197,86 @@ const char* ffDetectCPUImpl(const FFCPUOptions* options, FFCPUResult* cpu)
...
+            // In Asahi Linux, reading /proc/device-tree/compatible gives
+            // information on the device model. It consists of 3 NUL terminated
+            // strings, the second of which gives the actual SoC model. But it
+            // is not the marketing name, i.e. for M2 there is "apple,t8112" in
+            // the compatible string.
+            //
+            // A full list of the SoC model names can be found here:
+            // https://github.com/AsahiLinux/docs/wiki/Codenames
+            #define DT_COMPAT_PATH "/proc/device-tree/compatible"
+            if (strcmp(cpu->vendor.chars, "Apple") == 0)
```
CarterLi, line 209: ffStrbufEqualS(&cpu->vendor, "Apple")

CarterLi: Where do you get the vendor name?

src/detection/cpu/cpu_linux.c
```diff
@@ -197,6 +197,86 @@ const char* ffDetectCPUImpl(const FFCPUOptions* options, FFCPUResult* cpu)
...
 
+            // In Asahi Linux, reading /proc/device-tree/compatible gives
+            // information on the device model. It consists of 3 NUL terminated
+            // strings, the second of which gives the actual SoC model. But it
+            // is not the marketing name, i.e. for M2 there is "apple,t8112" in
+            // the compatible string.
+            //
+            // A full list of the SoC model names can be found here:
+            // https://github.com/AsahiLinux/docs/wiki/Codenames
+            #define DT_COMPAT_PATH "/proc/device-tree/compatible"
```
CarterLi, line 208: Maybe we should guard this with `#ifdef __aarch64__`

src/detection/cpu/cpu_linux.c
ydalton: Yep, that's an oversight on my part. It should be "M3 Max".

src/detection/cpu/cpu_linux.c
ydalton: Sure.

src/detection/cpu/cpu_linux.c
ydalton: The section above that runs `lscpu` and parses the  "Vendor ID" section correctly sets the vendor to "Apple". On my machine it looks like this:
```
Vendor ID:              Apple
  Model name:           Blizzard-M2
    Model:              0
```
I'll update to use the correct function.

PR #715, opened 2024-02-11
Add installation instructions for even more Linux distributions

AlbydST:
Added Alpine (now with the right command), ALT Linux and Arch AUR to the installation instructions.


CarterLi:
Thanks

PR #716, opened 2024-02-12
README: add MacPorts install info

herbygillot:
Add info on installing `fastfetch` on macOS via MacPorts: https://ports.macports.org/port/fastfetch/


CarterLi:
Thanks

PR #770, opened 2024-04-02
Update flashfetch.c

kojq:
add the skipped "host" option from "Flashfetch: simplify implementation" (in #645)

### DIFF
--- a/src/flashfetch.c
+++ b/src/flashfetch.c
@@ -21,6 +21,7 @@ int main(void)
         &options->title,
         &options->separator,
         &options->os,
+        &options->host,
         &options->kernel,
         &options->uptime,
         &options->packages,


CarterLi:
Thanks. May I ask why do you want to use `flashfetch`? They make no differences if you don't create a config file manually.

CarterLi:
Quick benchmark:

```
$ hyperfine -w 3 './fastfetch -c none' './flashfetch'
Benchmark 1: ./fastfetch -c none
  Time (mean ± σ):     210.7 ms ±  26.0 ms    [User: 4.4 ms, System: 14.0 ms]
  Range (min … max):   164.7 ms … 266.3 ms    14 runs

Benchmark 2: ./flashfetch
  Time (mean ± σ):     217.9 ms ±  25.6 ms    [User: 3.9 ms, System: 16.6 ms]
  Range (min … max):   170.8 ms … 259.2 ms    12 runs

Summary
  ./fastfetch -c none ran
    1.03 ± 0.18 times faster than ./flashfetch
```

kojq:
> Thanks. May I ask why do you want to use `flashfetch`? They make no differences if you don't create a config file manually.

I know, but placebo xD. There is little need for `flashfetch` (especially on a binary distro) since when I feel like using hefty config, it will make sense for me to not use `flashfetch` unless I maintain with constant compiling.

However, `flashfetch` comes compiled already... Shouldn't it make sense to be behind some configuration like a build flag so distros don't ship a practically pointless binary?

PR #772, opened 2024-04-03
fastfetch: add the missing new-line in config files

apocelipes:
It is better to end the text file with a newline character, otherwise some command line tools cannot correctly process the contents of the file. Here's a more detailed explanation: <https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline>

Also add some TODOs to remind me to replace these ugly `putc` with `YYJSON_WRITE_NEWLINE_AT_END` after a new version of `yyjson` is released.


src/fastfetch.c
```diff
@@ -840,25 +841,21 @@ static void writeConfigFile(FFdata* data, const FFstrbuf* filename)
     ffOptionsGenerateLibraryJsonConfig(&instance.config.library, doc);
     ffMigrateCommandOptionToJsonc(data, doc);
 
-    if (ffStrbufEqualS(filename, "-"))
-        yyjson_mut_write_fp(stdout, doc, YYJSON_WRITE_INF_AND_NAN_AS_NULL | YYJSON_WRITE_PRETTY_TWO_SPACES, NULL, NULL);
-    else
+    FF_AUTO_CLOSE_FILE FILE *fp = stdout;
```
CarterLi, line 844: Seems you are attempting to close stdout. Don't do that.

apocelipes: Done.

PR #792, opened 2024-04-13
Added screenshots of presets 6 to 12.

HyperHypa:
## Summary (Side rant)
I was initially looking through the presets and was confused about why there weren't any screenshots for presets 6 to 12, so I took some time to take a few pictures with Fastfetch on an unmodified terminal. I want to point out that not all the modules are shown since I would like not to have some of the information displayed publicly. I was also thinking about taking screenshots of the other presets so they could be used to create something consistent to compare, although that's something for later. On a side note, adding the pictures to the README.md would make it more convenient for others to see which preset they wish to build upon.

## TLDR
1. Added screenshots of presets 6 to 12 on an unmodified terminal.
2. Would recommend adding reference pictures to the customization section in the README.md.


CarterLi:
The icons of CPU & GPU seem not right.

HyperHypa:
Is it not a nerdfont icon?

CarterLi:
Seems some icons were removed in later version of nerd fonts

PR #817, opened 2024-04-28
Speedup nixos package count

xoltia:
I have made a couple of changes to the package counting for Nix.
- Instead of using a shell script, I reimplemented it in C which gives me the same results in slightly lower time.
- We can use `nix-store --query --hash` to check if there have been any changes before using query to count the packages. This is much faster when the cache is a hit.

Cached time vs time to query normally:

```
Benchmark 1: ./fastfetch
  Time (mean ± σ):      67.7 ms ±   2.2 ms    [User: 27.2 ms, System: 30.4 ms]
  Range (min … max):    63.5 ms …  72.3 ms    43 runs
 
Benchmark 2: ../../fastfetch-cli/fastfetch/build/fastfetch
  Time (mean ± σ):     323.7 ms ±  11.0 ms    [User: 125.5 ms, System: 122.8 ms]
  Range (min … max):   311.8 ms … 341.4 ms    10 runs
 
Summary
  ./fastfetch ran
    4.78 ± 0.22 times faster than ../../fastfetch-cli/fastfetch/build/fastfetch
```

### DIFF
--- a/src/detection/packages/packages_linux.c
+++ b/src/detection/packages/packages_linux.c
@@ -118,28 +118,152 @@ static uint32_t countFilesRecursive(FFstrbuf* baseDir, const char* dirname, cons
     return sum;
 }
 
+static bool isValidNixPkg(FFstrbuf* pkg)
+{
+    if (!ffPathExists(pkg->chars, FF_PATHTYPE_DIRECTORY))
+        return false;
+
+    ffStrbufSubstrAfterLastC(pkg, '/');
+    if (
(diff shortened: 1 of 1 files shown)


CarterLi:
I don't use nix. I just trust your code

nonetrix:
Great work, will be switching over to Fastfetch because of this just hoping it gets merged into nixpkgs unstable soon. 30ms does still seem a little slow compared to Arch for example but MUCH better than it was before, but is the any more room to reduce it further? I wish I could just cache the result and only update it when the system is rebuilt to be honest

PR #825, opened 2024-04-30
Extended Source Mage support

vaygr:
- correct Source Mage detection
- Source Mage package manager support

### DIFF
--- a/src/logo/builtin.c
+++ b/src/logo/builtin.c
@@ -3791,14 +3791,14 @@ static const FFlogo S[] = {
             FF_COLOR_FG_WHITE,
         },
     },
-    // SourceMage
+    // Source Mage
     {
-        .names = {"Source Mage", "source_mage"},
+        .names = {"Source Mage", "Source Mage GNU/Linux", "source_mage", "sourcemage"},
         .lines = FASTFETCH_DATATEXT_LOGO_SOURCE_MAGE,
         .colors = {
             FF_COLOR_FG_WHITE,
         },
-        .colorKeys = FF_COLOR_FG_BLUE,
+        .colorKeys = FF_COLOR_FG_RED,
         .colorTitle = FF_COLOR_FG_WHITE,
     },
     // Solaris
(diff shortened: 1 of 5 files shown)


src/modules/packages/packages.c
```diff
@@ -400,6 +406,7 @@ void ffPrintPackagesHelpFormat(void)
         "Number of winget packages",
         "Number of opkg packages",
         "Number of am packages",
+        "Number of spells",
```
CarterLi, line 409: spells?

src/logo/builtin.c
```diff
@@ -3791,14 +3791,14 @@ static const FFlogo S[] = {
             FF_COLOR_FG_WHITE,
         },
     },
-    // SourceMage
+    // Source Mage
```
CarterLi, line 3794: Please don't add the white space here

vaygr: @CarterLi is this breaking something? This is proper spelling for the distribution. See [Intro](https://sourcemage.org/Intro).

src/modules/packages/packages.c
vaygr: This is as per official [Glossary](https://sourcemage.org/Glossary).

CarterLi: The package manager name should be used because it's more intuitive, or brew will be called as formulae too

src/logo/builtin.c
CarterLi: No, just follow the coding style

src/modules/packages/packages.c
vaygr: Yeah, makes sense. Corrected.

src/logo/builtin.c
vaygr: @CarterLi not sure what you mean. The distribution is `Source Mage`, and not `SourceMage`, similar to `Chimera Linux` above. Should it be spelled `SourceMage` in this section anyway?

PR #841, opened 2024-05-02
Consistent writing style attempt

edi386:
I've wrote the extended description in the commit(s) before, to sum it up:

I have considered making an effort in having the writing be more consistent and readable. At the same time, I left the program's logic alone.

Letting the code have a mix of writing conventions, tight packing and inconsistencies would be detrimental for reading, especially when it comes to code review.

Instead, I suggest that the code has to follow one contemporary writing style, and all maintainers must adhere to it.

For my correction of the code, I have used both documentations:
- CS50 Style Guide for C - https://cs50.readthedocs.io/style/c/ (mostly)
- Linux kernel coding style - https://www.kernel.org/doc/html/v4.10/process/coding-style.html (a bit)

I would recommend the usage of CS50 one, as new C programmers will have likely went through the CS50 course, and the documentation is simple & on point (albeit somewhat vague).

### DIFF
--- a/src/fastfetch.c
+++ b/src/fastfetch.c
@@ -15,18 +15,24 @@
     #include "util/windows/getline.h"
 #endif
 
-static void printCommandFormatHelp(const char* command)
+static void printCommandFormatHelp(const char *command)
 {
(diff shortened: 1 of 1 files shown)


CarterLi:
I don't like this. Every project has its own coding style and they don't have to be the same. If there is inconsistant, just fix it instead of making it something different.

PR #857, opened 2024-05-03
Update README.md again for same reason different location

AxeldeWater:
Oversaw a place where it wasn't updated either


AxeldeWater:
Another update to the readme about the installation on Ubuntu/Debian pointed out by @D3SOX

PR #881, opened 2024-05-05
remove redundant checks

Nour1248:
we already check if value is equal to NULL in ffStrbufAppendNS


apocelipes:
The checks are not redundant because `value` and its data will be used as parameters. Skipping the check leds to UBs.
`ffStrbufAppendNS`'s check is also not redundant because `value`'s data could be NULL or empty.

PR #902, opened 2024-05-07
Free all allocated memory before exiting

apocelipes:
I always get a lot of noise when using memory detecting tools. Most of the noise comes from static variables that called malloc but did not freed. These leaks might be harmless but it's better to clean up all resources when we exit.

I split this into 4 commits so that when some parts run into problems it can be easily reverted.

### DIFF
--- a/src/detection/displayserver/displayserver.c
+++ b/src/detection/displayserver/displayserver.c
@@ -33,9 +33,31 @@ bool ffdsAppendDisplay(
 
 void ffConnectDisplayServerImpl(FFDisplayServerResult* ds);
 
+
+static inline void ffDisplayServerResultDestory(FFDisplayServerResult* result)
+{
+    if (!result) return;
+    ffStrbufDestroy(&result->wmProcessName);
+    ffStrbufDestroy(&result->wmPrettyName);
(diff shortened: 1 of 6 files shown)


CarterLi:
Memory assigned to static variables should not be counted as memory leak, because they are still accessable. GC wont collect them too.

Your change increase the code comlexity for completely no benefits. I dont think I like them

apocelipes:
> Memory assigned to static variables should not be counted as memory leak, because they are still accessable. GC wont collect them too.
That's not true. In C++ static storage objects will be destoryed before exiting. Currently there are only memory leaks, if there's a file handler the data could be lost.

The complexity is only added some destructors and nothing happens to APIs. And I use the cycle complexity tool and it reports the complexity is 9005, before this change it's 9003, nearly no impact on readability

apocelipes:
And you can take a look at this: [link](https://stackoverflow.com/questions/39385520/do-i-have-to-free-static-dynamically-allocated-pointers)

CarterLi:
image

apocelipes:
image

At the end of the process, this global variables will be not accessable.

PR #909, opened 2024-05-09
Terminal (macOS): fix compatibility with `(*term)`

5ouma:
Ignore `<shell> (<char>term)`.
This is because Fig(`<shell> (figterm)`) migrated to CodeWhisperer(`<shell> (cwterm)`) and CodeWhisperer renamed to Q(`<shell> (qterm)`).
Future changes will all be ignored by this change.
Related: #676


CarterLi:
What about ignore all ` (XXXterm)`?

5ouma:
How do you think the code should be?
Is this OK for you?
```c
if (ffStrContains(name, " (") && ffStrEndsWith(name, "term)")
```

CarterLi:
I think it's too loose. Can we ensure that there's only alpha chars between the parentheses?

5ouma:
What do you think about this?

```c
// terminalshell_linux.c

#ifdef __APPLE__
    // https://github.com/fastfetch-cli/fastfetch/discussions/501
    char n[256];
    strcpy(n, name);
    char *substring;
    substring = strtok(n, "(");
    substring=strtok(NULL, "term)");
    if (substring != NULL && ffStrIsEnglishAlphabet(substring))
    {
        if (__builtin_expect(getProcessNameAndPpid(ppid, name, &ppid, NULL) != NULL, false))
            return 0;
    }
#endif
```

```c
// stringUtils.h

static inline bool ffStrIsEnglishAlphabet(const char* str)
{
    for (size_t i = 0; i < strlen(str); i++)
        if (!ffCharIsEnglishAlphabet(str[i]))
            return false;
    return true;
}
```

CarterLi:
Try this?

```
// https://github.com/fastfetch-cli/fastfetch/discussions/501
{
    const char* pLeft = strstr(name, " (");
    if (pLeft)
    {
        pLeft += 2;
        const char* pRight = strstr(pLeft, 'term)');
        if (pRight && pRight[5] == '\0')
        {
            for (; pLeft < pRight; ++pLeft)
            {
                if (*pLeft < 'a' || *pLeft > 'z')
                    break;
            }
            if (pLeft == pRight && getProcessNameAndPpid(ppid, name, &ppid, NULL) != NULL)
                return;
        }
    }
}
```

PR #910, opened 2024-05-09
JsonConfig: Added `brightness` properties and `dim_` colors

djinnalexio:
I was making a custom config in VScode and noticed issues with Intellisense.

First, most of the properties for the `brightness` module weren't valid according to the template.

Second, when using colors, I noticed that `bright_` colors are returning the same color as the base ones. I then tried with the `dim_` prefix which did give me the other shades which were the darker ones.  I got this behavior on both Fedora 40 GNOME 46 and Ubuntu 22.04 GNOME 42.9.  I am not sure if the issue is GNOME or Linux-specific which is why I only added `dim_` variants to the color list instead of replacing the `bright_` ones.

### DIFF
--- a/doc/json_schema.json
+++ b/doc/json_schema.json
@@ -1003,6 +1003,21 @@
                                         "minimum": 0,
                                         "maximum": 400,
                                         "default": 10
+                                    },
+                                    "key": {
+                                        "$ref": "#/$defs/key"
+                                    },
+                                    "keyColor": {
+                                        "$ref": "#/$defs/keyColor"
+                                    },
+                                    "outputColor": {
+                                        "$ref": "#/$defs/outputColor"
+                                    },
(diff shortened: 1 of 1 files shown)


CarterLi:
`dim_` `bright_` are meant to be used by combining standard color keywords. Adding all the combinations will make the list too verbose.

In addition, 256 color and true color can also be specified. See `fastfetch -h color` for detail. JSON schema is main for suggestion purpose, not validation.

> I noticed that bright_ colors are returning the same color as the base ones.

Some colors (like keys) are already in bright mode. Try `--bright-color false`

djinnalexio:
> In addition, 256 color and true color can also be specified. See fastfetch -h color for detail. JSON schema is main for suggestion purpose, not validation.

I get how it works now, thank you.

PR #918, opened 2024-05-13
SpoinkOS Logo

SpoinkOSDevs:


CarterLi:
It's too large. A good looking logo should not be higher than 20 rows.

PR #954, opened 2024-05-21
Added CuerdOS support

CuerdOS:
It looks really really great

image_2024-05-21_12-17-47


src/detection/os/os_linux.c
```diff
@@ -157,6 +157,13 @@ static bool detectDebianDerived(FFOSResult* result)
         ffStrbufSetNS(&result->versionID, versionEnd - versionStart, result->prettyName.chars + versionStart);
         return true;
     }
+    else if (ffStrbufStartsWithS(&result->name, "CuerdOS GNU/Linux"))
+    {
+        ffStrbufSetS(&result->name, "CuerdOS GNU/Linux");
+        ffStrbufSetS(&result->id, "cuerdos");
+        ffStrbufSetS(&result->idLike, "debian");
```
CarterLi, line 164: Doesn't your os-release file provide these info?

CuerdOS: Yes, this is how we have it set up

imagen

CarterLi: So don't hard code these info. Remove these `ffStrbufSetS`s

CuerdOS: Did it

Issue #970, opened 2024-05-27
[FEAT] Add support for guix package manager


Dariqq:
## Wanted features:
Guix is a package manager similiar to the nix package manager.
Currently packages installed via guix are not discovered. 

There are 3 main profiles where packages are installed:

- `/run/current-system/profile`: Packages installed system wide  (if running Guix System)
- `~/.guix-profile`: Packages installed by the current user
- `~/.guix-home/profile`: Packages installed by the current user with Guix Home

One can query the packages installed in a profile with `guix package -p PROFILE -I`  however it only lists packages that are explicitly installed.

You can also list all requisites of a profile with `guix gc -R $(realpath PROFILE)`  but that output would need some parsing as not all entries correspond to a package (and is quite slow).  again as nix package manager this still misses some some packages (like the kernel) but its a at least a better approximation of package count as the first option.

I started trying to implement things myself but got a bit lost where all the new counts from `FFPackagesResult` would need to be added to make things work as expected. 

## Motivation:

Currently packages installed via guix are not discovered. 
Neofetch and other neofetch-like programs do show package counts for guix.

CarterLi:
> I started trying to implement things myself but got a bit lost where all the new counts from `FFPackagesResult` would need to be added to make things work as expected.

```diff
diff --git a/src/detection/packages/packages.h b/src/detection/packages/packages.h
index 4e954b96..0c6cf33c 100644
--- a/src/detection/packages/packages.h
+++ b/src/detection/packages/packages.h
@@ -31,6 +31,8 @@ typedef struct FFPackagesResult
     uint32_t sorcery;
     uint32_t winget;
     uint32_t xbps;
+    uint32_t guixSystem;
+    uint32_t guixUser;
 
     uint32_t all; //Make sure this goes last
```

Then add the logic in `packages_linux.c`. You may do the basic stuff and I can have it polished.

PR #972, opened 2024-05-27
Packages (Linux): Add Guix package manager.

Dariqq:
Add the 3 main profiles for guix package manager.

The detection is based on the one for neofetch and only includes packages explicitly installed by the user or propagated from other packages. 

I have not yet found a good way to implement a more sophisticated approach (i.e. something similar as what is done for nix package manager). Should this be in a second merge request once I have it ready, but when I have more time to look more deeply into it?

Hopefully I caught all position where to include the new package manager option in modules/packages/packages.c.

Also had some trouble with whitespace before in the commit, hopefully i caught it all.

I have built the package with this commit successfully on x64-linux-gnu and aarch64-linux-gnu (via qemu-binfmt).

Tested it on a sytem running Guix System and on Fedora running guix as additional package manager.

Closes https://github.com/fastfetch-cli/fastfetch/issues/970

### DIFF
--- a/src/detection/packages/packages.h
+++ b/src/detection/packages/packages.h
@@ -14,6 +14,9 @@ typedef struct FFPackagesResult
     uint32_t eopkg;
     uint32_t flatpakSystem;
     uint32_t flatpakUser;
+    uint32_t guixHome;
+    uint32_t guixSystem;
+    uint32_t guixUser;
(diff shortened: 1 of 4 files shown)


CarterLi:
I think it's pretty solid. LGTM

PR #981, opened 2024-05-30
Add ability to skip installing license with INSTALL_LICENSE option

Maytha8:
This is useful for Debian, where the package already includes copyright information and therefore installing the license into /usr/share/licenses is not needed.

Upstream support for an option to disable installing the license is much better than patching it on the packaging side (which needs to be continuously updated).


CarterLi:
Thanks. How is the debian packaging?

Issue #989, opened 2024-06-04
[FEAT] Added newline after modules print


retrobit:
Be sure to read [FAQ](https://github.com/fastfetch-cli/fastfetch?tab=readme-ov-file#faq) before submitting a new issue.

## Wanted features:

Simply put, a newline to separate the fetch output from any shell greeting or the prompt would be nice. I figure it would be best in `flashfetch.c` after module printing, or in the call to `init.c`s `ffFinish()` right after.

## Motivation:

It's purely subjective, but Neofetch had it (maybe two, actually) and it looks nice/it's easier to read. Examples of Neofetch output vs fastfetch output attached.

![fastfetch](https://github.com/fastfetch-cli/fastfetch/assets/6226450/7415f96c-a4ab-405f-beeb-e6ee8c6ed63c)
![Neofetch](https://github.com/fastfetch-cli/fastfetch/assets/6226450/42098027-73f7-497f-a9c9-54d8a4ebeb0e)

CarterLi:
Two options:

1. Add a `"break"` at the end of `config.jsonc`
2. Use `fastfetch && echo`

retrobit:
Thanks! I hadn't considered the configuration file solution; I did want to avoid an `echo` as it seemed less clean.

retrobit:
Configuration solution worked perfectly

PR #990, opened 2024-06-04
Added newline after modules print

retrobit:
World's simplest pull request...

Feature enhancement issue was created by me that has all the details: https://github.com/fastfetch-cli/fastfetch/issues/989

Not sure why, but the newline was not working in `flashfetch.c` after module printing, but worked in `init.c`s `ffFinish()` method.

Built and tested on macOS, attached is a screenshot with the behavior showing the newline: 
fastfetch after change

### DIFF
--- a/src/common/init.c
+++ b/src/common/init.c
@@ -130,6 +130,9 @@ void ffFinish(void)
     if(instance.config.logo.printRemaining)
         ffLogoPrintRemaining();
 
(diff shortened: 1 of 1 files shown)


CarterLi:
No. There are many ways to print a newline. However if you hard coded it into fastfetch, it's not easy to remove it.

retrobit:
Valid, as it's preferential and someone might want it to run up against their shell greeting or prompt.

PR #999, opened 2024-06-06
Packages (Linux): Improve speed of Guix package detection.

### LINKED ISSUE #988: [FEAT]  Make guix package manager detection faster
While putting fastfetch 2.14 on a slower laptop of mine i noticed that my implementation for package counts for guix is quite slow taking 60-70ms for each of the 3 profiles on my main machine and up to 2-3x that on a slow laptop.

The main cause for this is waiting for guile to return the list of installed packages. 

I have taken a look at what is happening behind the scenes: 
- The `-I` flag of `guix package` calls the `list-installed` function from [here](https://git.savannah.gnu.org/cgit/guix.git/tree/guix/scripts/package.scm#n777).
- Which then uses `profile-manifest` (from [here](https://git.savannah.gnu.org/cgit/guix.git/tree/guix/profiles.scm#n284)) to read and parse the manifest of the profile at `PROFILE/manifest`.

This manifest files currently look like this:

```scm
(manifest
 (version 4)
 (packages
  (("emacs"
    "29.3"
    "out"
    "/gnu/store/1bi7n031g6b5b07hvps3gyi0rhi44qkb-emacs-29.3"
    ;; ...
    )
   ("fastfetch"
    "2.14.0"
    "out"
    "/gnu/store/kamkxrxi8mdcfagfg0z4nfranhfn8csb-fastfetch-2.14.0"
    ;; ...
    )
   ("btop"
    "1.3.2"
    "out"
    "/gnu/store/0j1ya0g65nk613lnq56dyhnv4fk8blvb-btop-1.3.2")
   ;; .....
   )))
```
(Somtimes there is additional information for a package like search paths, propagated packages or provenance information, etc which I have omitted here for clarity)

All that gets displayed using `guix package -I` are the entries in the `packages` list. 

Would it be possible to parse this file with C instead? Are there any other examples I could look at which do something similiar?

EDIT: 
The thing that I tried was using `GetNumStrings` for `"/gnu/store/"` but that also catches additional entries from the additional fields that i omitted above. But at least it is a lot faster reducing the total time for packages to 4ms

Dariqq:
Here is logic of your first code. There was not much I needed to do.

Compared to it I've renamed the argument to `filename` to be more explicit what it acts on, have replaced the last  `memcmp` call with `compare32` and added the comment from my first try to try to explain what the code should do. 

Thanks for helping me with this, I have added you as Co-author because you did almost everything.

Haven't tested how it interacts which edge cases (like malformed manifest files) as these are symlinked from a read-only filesystem and i figured if this happens problems with fastfetch are probably not the top priority.

Speed wise it is now much faster from the 200-300ms total to ~5ms and the count is at a lot more accurate (but still not perfect) as (transitive) dependencies of installed packages are now also counted. 

Closes #988.

### DIFF
--- a/src/detection/packages/packages_linux.c
+++ b/src/detection/packages/packages_linux.c
@@ -408,26 +408,36 @@ static uint32_t getAM(FFstrbuf* baseDir)
     return result;
 }
 
-
-static uint32_t getGuixPackagesImpl(char* path)
+static int compareHash(const void* a, const void* b)
 {
-    FF_STRBUF_AUTO_DESTROY output = ffStrbufCreateA(1024);
+    return memcmp(a, b, 32);
+}
 
-    ffProcessAppendStdOut(&output, (char* const[]) {
-        "guix",
-        "package",
-        "-p",
-        path,
-        "-I",
-        NULL
-    });
+static uint32_t getGuixPackagesImpl(char* filename)
+{
+    FF_STRBUF_AUTO_DESTROY content = ffStrbufCreate();
+    if (!ffAppendFileBuffer(filename, &content))
+        return 0;
 
+    // Count number of unique /gnu/store/ paths in PROFILE/manifest based on their hash value.
+    // Contains packages explicitly installed and their propagated inputs.
+    char* pend = content.chars;
 
-    //Each package is a new line in the output.
-    // If at least one line is found, add 1 for the last line.
-    uint32_t count = ffStrbufCountC(&output, '\n');
-    if(count > 0)
-      count++;
+    for (const char* pattern = content.chars; (pattern = strstr(pattern, "/gnu/store/")); pattern += 32)
+    {
+        pattern += strlen("/gnu/store/");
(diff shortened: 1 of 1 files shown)


CarterLi:
Do you know how can I submit fastfetch to GNU guix repo?

CarterLi:
What does `lurk` or `strace` print?

CarterLi:
Just saw someone submitted fastfetch to guix repo with a few features enabled but no `LD_LIBRARY_PATH` set

https://git.savannah.gnu.org/cgit/guix.git/commit/?id=ddb0401d82be8b1a43b6f47fb5794d08119f8d54

Does it really work?

CarterLi:
I don't like the idea. You shouldn't force people to install wayland if they use only x11.

One solution is to set rpath in `-DCMAKE_EXE_LINKER_FLAGS`. See https://github.com/fastfetch-cli/fastfetch/blob/dev/CMakeLists.txt#L871-L873 for reference

PR #1027, opened 2024-06-19
Add the missing '--users-myself-only' option in bash completions

Antiz96:
The new `--users-myself-only` option released in `v2.16.0` is missing from the bash completions.

### DIFF
--- a/completions/fastfetch.bash
+++ b/completions/fastfetch.bash
@@ -347,6 +347,7 @@ __fastfetch_completion()
         "--users-key"
         "--users-format"
         "--users-key-color"
+        "--users-myself-only"
         "--bluetooth-key"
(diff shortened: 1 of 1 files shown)


CarterLi:
The bash completions file is very outdated and it lacks a lot more flags. It should use `fastfetch --help-raw` as what fish completions do.

I'm not good at writting shell scripts. If someone would like to contribute I'm happy to provide some support.

Antiz96:
Oh okay, I didn't notice the file was outdated and missing more flags, my bad... Thanks for the carapace-bin pointer :)

CarterLi:
Before the script gets rewritten, adding any missing flags is still useful. Thanks for this PR!

Issue #567, opened 2023-09-27
[linux] Can't recognize the cpu name correctly on arm64.


xzl01:
Sorry I didn't follow your feedback format. I noticed that on Huawei Pangu ARM machines the CPU name is output as unknown:
![image](https://github.com/fastfetch-cli/fastfetch/assets/31403725/f727d559-1c50-4355-8521-bb26d69e66a9)
But the name output on the lscpu is correct, After looking at the code I realized that the lscpu does something special for arm64 cpus  <https://github.com/util-linux/util-linux/blob/87f491671efb920ddd1c45c70dc8016f072851bb/sys-utils/lscpu-arm.c> 

Hopefully fastfetch will also support recognizing

CarterLi:
I will just call lscpu and fetch its output

daniejstriata:
I have output from the AWS g4 lscpu in this ticket: https://github.com/fastfetch-cli/fastfetch/issues/566

CarterLi:
What's `lscpu -p=MODELNAME`?

CarterLi:
![Screenshot_2023-09-27-14-16-19-753_com.termux.jpg](https://github.com/fastfetch-cli/fastfetch/assets/6134068/b8f0daa2-06bd-47ae-bc35-8c061cd31f44)

Tested on Android, lscpu prints different model names for different cores. In this case should fastfetch use only the first name, or print them all?

daniejstriata:
`lscpu | awk '/Model name:/ { print $3 }'`
Neoverse-N1

CarterLi:
What's `lscpu -h`?

> lscpu | awk '/Model name:/ { print $3 }'

I can't use that, because the `Model name` is localized.

CarterLi:
Seems that `CPU part` is `id_part::id` in `lscpu-arm.c`. I decided to use it.

xzl01:
This is the output of my lscpu:
```
deepin@deepin-pc-52:~$ lscpu
Architecture:           aarch64
  CPU op-mode(s):       64-bit
  Byte Order:           Little Endian
CPU(s):                 24
  On-line CPU(s) list:  0-23
Vendor ID:              HiSilicon
  Model name:           Kunpeng-920
    Model:              0
    Thread(s) per core: 1
    Core(s) per socket: 24
    Socket(s):          1
    Stepping:           0x1
    BogoMIPS:           200.00
[...]

PR #1033, opened 2024-06-22
prioritize lscpu model output for arm linux systems if available

theofficialgman:
resolves https://github.com/fastfetch-cli/fastfetch/issues/567#issuecomment-2183677298

### DIFF
--- a/src/detection/cpu/cpu_linux.c
+++ b/src/detection/cpu/cpu_linux.c
@@ -298,75 +298,142 @@ const char* ffDetectCPUImpl(const FFCPUOptions* options, FFCPUResult* cpu)
     detectAsahi(cpu);
     #endif
 
-    if (cpu->name.length == 0)
+    #if defined(__aarch64__) || defined(__arm__) //prioritize lscpu on arm for better model name output
+    FF_STRBUF_AUTO_DESTROY buffer = ffStrbufCreate();
+    if (ffProcessAppendStdOut(&buffer, (char *const[]) { "lscpu", NULL }) == NULL)
     {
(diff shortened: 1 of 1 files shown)


CarterLi:
It's not fallback. It's force.

lscpu should always used as fallback, because you cant assume it's available.

theofficialgman:
> It's not fallback. It's force.
> 
> lscpu should always used as fallback, because you cant assume it's available.

corrected name.

feel free to rework so that the lscpu section is called first so that it takes priority rather than forcing it.

theofficialgman:
@CarterLi how about something like that ^
Now lscpu is only prioritized and will fallback to /proc/cpuinfo if lscpu is not available

CarterLi:
I dont like the idea. The CPU name was reported by Linux kernel. If you think it is wrong, it is a kernel issue.

theofficialgman:
> I dont like the idea. The CPU name was reported by Linux kernel. If you think it is wrong, it is a kernel issue.

its not an issue. same deal with apple silicon which you have codepaths for. this is just how ARM cpu's have always been represented (with their implementer, architecture, variant, and part numbers) with no "pretty" model name

https://github.com/util-linux/util-linux/blob/master/sys-utils/lscpu-arm.c

theofficialgman:
also with this you can remove the hacky asahi codepaths

theofficialgman:
https://github.com/util-linux/util-linux/commit/744d62ee0c54963539832ec5943f3d25e0fccfbd

image

so yeah sorry not going to happen. just use lscpu like everyone else

CarterLi:
Won't fix

PR #1038, opened 2024-06-24
OS (Linux): prioritize `prettyName` over `name`

xbt573:
PR title pretty much says all about changes in code. Currently fastfetch uses `os-release`'s `name` over `prettyName`, which are not always equal (for example Gentoo's `name` is `Gentoo`, but `prettyName` is `Gentoo Linux`)

This change just swaps order of `name` and `prettyName` in if-else if-else chain

### DIFF
--- a/src/modules/os/os.c
+++ b/src/modules/os/os.c
@@ -12,10 +12,10 @@
 static void buildOutputDefault(const FFOSResult* os, FFstrbuf* result)
 {
     //Create the basic output
-    if(os->name.length > 0)
-        ffStrbufAppend(result, &os->name);
-    else if(os->prettyName.length > 0)
+    if(os->prettyName.length > 0)
         ffStrbufAppend(result, &os->prettyName);
+    else if(os->name.length > 0)
+        ffStrbufAppend(result, &os->name);
     else if(os->id.length > 0)
         ffStrbufAppend(result, &os->id);
     else


CarterLi:
This PR is known to break alpine linux and debian.

The pretty name of apline linux is `Alpine Linux v3.20` while its version is `3.20.0`, which results in

```
OS: Alpine Linux v3.20 3.20.0 aarch64
```

It may also break debian. We fetch point release version of debian in `/etc/debian_version`, while pretty name of debian contains only the major version. so that fastfetch will print `Debian GNU/Linux 12 (bookworm) 12.1 x86_64`. See https://github.com/dylanaraps/neofetch/issues/2381

CarterLi:
Unless it gets tested in every linux distro, I want the old behavior being kept

xbt573:
I tested some popular distros with docker (and my own Gentoo), and i can categorize the results into some categories:
1. Everything is okay, `name` equals `prettyName`: Arch Linux, Clear Linux
2. Everything is okay, `prettyName` contains `version_id` (version duplication prevented by condition): Fedora, RockyLinux
3. Patch did output better: Alt Linux (Alt `name` is "starter kit"), Alma Linux (codename added, version duplication prevented by condition), Amazon Linux (version from `prettyName` is more verbose), Gentoo Linux
4. Broken: Alpine Linux, Ubuntu Linux, Debian Linux

I propose these solutions:
1. Introduce some sort of exception list, and use `name` over `prettyName` if exception list contains `name`
2. Fuzzy version contains: Allow matching version in `prettyName` by parts, will help with `v3.20`-`3.20.0` and similar, harder
3. Don't include version if prettyName is used, radical but most of distros specify version in `prettyName` (8 of 11 tested, 2 is rolling)

I'm ready to assist with every solution you choose and test all fastfetch supported distros (by logo) if needed, if none of solutions is suitable then just close issue as "non planned"

CarterLi:
I would rather fix the `name` and `version`. 

For example: test if `prettyName` contains `name` + `" Linux"`

CarterLi:
Fixed in 6429c20

PR #1047, opened 2024-06-26
misc Qt improvements

capezotte:
First commit add supports for [Kvantum](https://github.com/tsujan/Kvantum/), a sort of meta-widgetstyle for Qt. When it's being used, it makes sense to source the style from Kvantum's configuration instead of just printing `kvantum`/`kvantum-dark`.

The second adds support for qt5ct and qt6ct, which are often used to configure Qt graphically in environments where Qt is not native.


CarterLi:
LGTM

PR #1076, opened 2024-07-07
fish completion: pipe straight to Python

zanchey:
In https://github.com/fish-shell/fish-shell/issues/10599, a user of an older version of fish has run into an issue with the fastfetch completion requiring a relatively new version.

Rewrite the (really quite clever) embedded Python into a straight pipe, dropping the string transformation entirely. Does require some escaping of quotes to support Python < 3.12.

### DIFF
--- a/completions/fastfetch.fish
+++ b/completions/fastfetch.fish
@@ -54,51 +54,48 @@ function __fastfetch_complete_structure
     end
 end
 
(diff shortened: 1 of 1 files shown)


CarterLi:
Thanks. I hope fish will support some bash-like syntax of embedding strings in future.

PR #1161, opened 2024-08-06
Support cpu detection and host info on PowerPC

barracuda156:
This adds CPU detection for PowerPC and support for human-friendly host names (otherwise model codenames are displayed).

Tested on 10.6 and 10.4.11, for PowerMac G5 Quad and PowerBook G4. Looks like everything works as expected:

fastfetch_tiger2

fastfetch_powerbook

### DIFF
--- a/src/detection/cpu/cpu.c
+++ b/src/detection/cpu/cpu.c
@@ -39,3 +39,14 @@ const char* ffCPUAppleCodeToName(uint32_t code)
         default: return "Apple Silicon";
     }
 }
+
+const char* ffCPUApplePPCCodeToName(int cpuSubType)
+{
+    switch (cpuSubType)
+    {
+        case 10: return "PowerPC 7400";
+        case 11: return "PowerPC 7450";
+        case 100: return "PowerPC 970";
+        default: return "PowerPC";
+    }
+}
--- a/src/detection/cpu/cpu.h
+++ b/src/detection/cpu/cpu.h
@@ -30,3 +30,4 @@ typedef struct FFCPUResult
 
 const char* ffDetectCPU(const FFCPUOptions* options, FFCPUResult* cpu);
 const char* ffCPUAppleCodeToName(uint32_t code);
+const char* ffCPUApplePPCCodeToName(int cpuSubType);
--- a/src/detection/cpu/cpu_apple.c
+++ b/src/detection/cpu/cpu_apple.c
@@ -100,12 +100,23 @@ static const char* detectCoreCount(FFCPUResult* cpu)
 
 const char* ffDetectCPUImpl(const FFCPUOptions* options, FFCPUResult* cpu)
 {
+#ifdef __POWERPC__
+    int cpuSubType;
+    size_t size = sizeof(cpuSubType);
+    if (sysctlbyname("hw.cpusubtype", &cpuSubType, &size, NULL, 0) == 0) {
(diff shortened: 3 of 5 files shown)


src/detection/cpu/cpu.c
```diff
@@ -24,6 +24,9 @@ const char* ffCPUAppleCodeToName(uint32_t code)
     // https://github.com/AsahiLinux/docs/wiki/Codenames
     switch (code)
     {
+        case 10: return "PowerPC 7400";
+        case 11: return "PowerPC 7450";
+        case 100: return "PowerPC 970";
```
CarterLi, line 29: This function is only for Apple Silicon code name conversion. Please don't mess it up.

CarterLi:
What's that weird leading `\` in mount points?

src/detection/cpu/cpu.c
barracuda156: I haven’t found any instances of its usage, perhaps missed it. Anyway, sure, I can make a separate one with PowerPC values.

barracuda156:
> What's that weird leading `\` in mount points?

@CarterLi Thanks for pointing out, I will look into why that is added. (We obviously need extra patches to run this on older macOS, so likely something went a bit wrong with that.)

src/detection/cpu/cpu.c
CarterLi: It's used for Asahi Linux, not for macOS

barracuda156: Ah, got it. I will remove ppc from here and make its on function for it.

CarterLi:
The refresh rate in the first screenshot is wrong too. I don't think you have a 40kHz display.

In the second screenshot, you have **10k+** macport packages installed?

CarterLi:
I think you should keep these patches locally. They don't magically make fastfetch work for PPC mac, neither will it improve fastfetch for intel and apple silicon. I think it's useless for fastfetch mainstream.

barracuda156:
> The refresh rate in the first screenshot is wrong too. I don't think you have a 40kHz display.

Should be fixable, but unrelated to this PR.

> In the second screenshot, you have **10k+** macport packages installed?

Yes, this is accurate :)

barracuda156:
@CarterLi I have fixed it according to your suggestion, so that not to mess up Asahi, and verified it builds and cpu and host detection work fine.

CarterLi:
I still don't think it's worth merging.

barracuda156:
@CarterLi It is up to you and other members of the project, of course, but if the concern is that nobody uses powerpc Macs anymore so it cannot benefit anymore, this is not the case, which you may see from subforum on [MacRumors](https://forums.macrumors.com/forums/powerpc-macs.145/), for example: it is pretty active.
Not too many people use MacPorts also, so it makes sense to have fixes with upstream rather that exclusive to MacPorts.

CarterLi:
Again, the problem is: even if I merged this PR, PPC mac still doesn't work.

All or nothing.

barracuda156:
@CarterLi If you help a bit with handling plists so that it doesn’t lead to segfaults (I _have_ a fix, but do not want to do it via macro-separated cases, if possible), we can probably make it ”all”.

Handling of modules to be disabled can be done via CMake option (`-DMAC_LEGACY` or alike).

I can see how to minimize the patches needed.

Can you take a brief look btw at what we have in MacPorts now? (Those are _not_ done in a compatible way, we apply them conditionally.)

PR #1165, opened 2024-08-06
Add support for Linux From Scratch and LFS-like build scripts

AlbydST:
Added support for:
- Linux From Scratch (LFS)
- Musl Linux From Scratch (MLFS)
- Clang-built Musl Linux From Scratch (CMLFS)
- Optimized Musl Linux From Scratch (Optimux)
- Other LFS build scripts (Yosild, Yamlld)

I think I've done everything correctly however from my testing the logo only displays with LFS, all the others just use the standard Linux logo so let me know if I've done anything wrong.


CarterLi:
You put `"mlfs", "cmlfs", "optimux", "yosild", "yamlld"` in `L` section, they won't work.

Issue #1197, opened 2024-08-18
[FEAT] Please add completion for zsh


AtifChy:
Be sure to read [FAQ](https://github.com/fastfetch-cli/fastfetch?tab=readme-ov-file#faq) before submitting a new issue.

## Wanted features:
`fastfetch` has a vast number of options available, it would be much easier to traverse around these options with tab completion instead of opening the man page over and over.

## Motivation:
currently fastfetch has completion support for `fish` shell with proper description beside each option, having the same completion with zsh would be awesome.

CarterLi:
You can try carapace-bin https://github.com/carapace-sh/carapace-bin

AtifChy:
@CarterLi i noticed that a python script being used to generate the completions for fish shell, can't the same be done for zsh. carapace-bin looks interesting, but i would prefer native shell completion.

CarterLi:
> i noticed that a python script being used to generate the completions for fish shell, can't the same be done for zsh

You still need to know the syntax of zsh script to generate the completions.

You can try it yourself if you have interest. `fastfetch --help-raw` prints a JSON document that contains all command line options. All you need to do is parse the JSON document [and generate zsh completion command](https://github.com/zsh-users/zsh-completions/blob/master/zsh-completions-howto.org)

aymen-3:
Since fastfetch has a --help option, you can add this line to your .zshrc file and it will have completion: 

**compdef _gnu_generic fastfetch**

Simon238:
For those of us using Oh-My-Zsh:

Copying [`fastfetch/completions/fastfetch.zsh`](https://github.com/fastfetch-cli/fastfetch/blob/dev/completions/fastfetch.zsh) to `~/.oh-my-zsh/cache/completions/_fastfetch` as is did not work for me. I had to change
```zsh
_fastfetch "$@"
```
into
```zsh
if [ "$funcstack[1]" = "_fastfetch" ]; then
    _fastfetch "$@"
else
    compdef _fastfetch fastfetch
fi
```

PR #1213, opened 2024-08-25
feat: add zsh-completion

AtifChy:
#1197 

This pull request introduces completion for zsh. I've customized the python script originally used to generate fish shell completion to generate zsh completion.


CarterLi:
How can I test it?

AtifChy:
> How can I test it?

`fpath=(repos/fastfetch/completions $fpath)`

you may need to also run `autoload -Uz compinit && compinit`

then run `fastfetch <TAB>`

CarterLi:
* `_fastfetch` should be renamed to `fastfetch.zsh`
* This file should be installed by cmake. See https://github.com/fastfetch-cli/fastfetch/blob/dev/CMakeLists.txt#L1340-L1344

AtifChy:
> * `_fastfetch` should be renamed to `fastfetch.zsh`
but this is how zsh completion files are named

CarterLi:
The file can be renamed when installing

AtifChy:
> * This file should be installed by cmake. See https://github.com/fastfetch-cli/fastfetch/blob/dev/CMakeLists.txt#L1340-L1344

done

CarterLi:
Thanks

PR #1216, opened 2024-08-26
fix isDir condition checking in io_unix.c

AtifChy:
the condition was failing to set `isDir` because of miss placed `else`. causing `fastfetch --list-presets` to include directories

image

here you can see `.` `..` and `examples` directory alongside other presets

with the fix:
image

i also updated zsh-completion to reflect upon this change


CarterLi:
`isDir` issue should be fixed in the dev branch. Please create a new PR to make the zsh completion related changes

PR #1221, opened 2024-08-28
some changes related to zsh-completion

AtifChy:
- [x] add completion for `--logo-color-{1..9}`
- [x] add completion support for passing config files with `--config`
- [x] some other improvement

### DIFF
--- a/completions/fastfetch.zsh
+++ b/completions/fastfetch.zsh
@@ -3,9 +3,8 @@
 function _fastfetch() {
   local state
 
-  local -a opts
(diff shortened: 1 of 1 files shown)


CarterLi:
I mean you may also take a look at the bash script.

AtifChy:
> I mean you may also take a look at the bash script.

sorry if this is bothering, i am constantly discovering these minor things. And for the bash script, it seems too verbose. I am only doing this because I am having fun with zsh shell scripting. 😅

PR #1225, opened 2024-08-29
Add Magix art for fastfetch

Paranoid-Pufferfish:
Adding the Ascii art to Magix Linux, a Gentoo based distro that is currently in development . 

Org repo : https://github.com/Magix-OS/

Screenshot: 
image

if there is something i have to add/fix, let me know ^^ And have a lovely day


CarterLi:
I know you use idea, but you should not commit the `.idea` stuff.

FabioLolix:
Commits should be squashed so the files don't get into the git history, no?

Paranoid-Pufferfish:
I think i will make a new repo and a new pull request will be better imo.

PR #1248, opened 2024-09-04
Fix grammer in README.md

yonas:


CarterLi:
I like grammer fixes. Thanks.

PR #1251, opened 2024-09-05
zsh completion: add file completion for `--logo`

AtifChy:
- [x] add file completion for `--logo` for custom logo
- [x] better description?


CarterLi:
> add file completion for --logo for custom logo

I don't like this. `--logo` should be used for builtin logo only. As for image logo, image protocol shortcuts should be used.

AtifChy:
ok

PR #1291, opened 2024-09-23
refactor: Improve memory metric parsing

nnyyxxxx:


CarterLi:
Can you clarify what is **improved**?

nnyyxxxx:
> Can you clarify what is **improved**?

this version uses a for-loop instead of multiple conditional checks

and other minor tweaks

CarterLi:
No. You made it slower. If `MemAvailable` is available, other properties should not be searched.

nnyyxxxx:
> No. You made it slower. If `MemAvailable` is available, other properties should not be searched.

would you like me to fix that :upside_down_face:

CarterLi:
No

PR #1292, opened 2024-09-24
Update fastfetch.1.in to have all the commands in the man page

### LINKED ISSUE #830: [FEAT] More documentation in the Manpage
## Wanted features:

- Include all the available options, currently only a subset (compared to `fastfetch --help`) is actually listed.
- Include the information from running `fastfetch -h <option>` for all options.
- Explain/give examples of the structure of the config file, accepted values for options, how to change common things, etc.
- Include the JSON Schema found in the github wiki.

## Motivation:

When I want to read thorough documentation for a tool I look at the manpage (as many others might do). I expect to find more information in `man foo` than running `foo --help`. Right now, some of the info is available in the manpage, more is in `fastfetch --help` and other is in the github wiki.

Having all the documentation in the manpage would be beneficial because:

1. There would be a centralized place to look for the documentation, without needing to look at external resources or having to run multiple commands (running `fastfetch -h <option>` constantly is not very ergonomic).
2. Users can leverage the features of the program they use to read manpages to search for specific topics of their interest more easily.

darwincereska:
I updated the man page to have all the commands from:
``` bash
fastfetch --help
```
and
``` bash
fastfetch -h [option]
```
Resolved #830

### DIFF
--- a/doc/fastfetch.1.in
+++ b/doc/fastfetch.1.in
@@ -23,7 +23,104 @@ the exit code will be non\-zero.
 
 .SH OPTIONS
 
-.SS "Informative Options"
+.SS "Informative options:"
+.HP
+\fB\-h\fR, \fB\-\-help\fR <?command>: Show this message, or help for a specific command
+.HP
+\fB\-v\fR, \fB\-\-version\fR: Show the full version of fastfetch
+.HP
+\fB\-\-version\-raw\fR: Show the raw version string (major.minor.patch)
+.HP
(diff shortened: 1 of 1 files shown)


CarterLi:
1. Seems you are using a very old version of fastfetch. Some options are missing; some options have been deprecated and removed.
2. Seems you generate it with some 3rd party tool. It needs to be polished.
3. You need to provide instructions to maintain it, unless you want to do it in every future release.

image
image

PR #1301, opened 2024-09-28
Display (Linux): detect all enabled or connected connectors

ziyao233:
Background: I found fastfetch on one of my ARM board cannot display CVBS port related information.

DRM connectors have two properties, "enabled" and "status", standing for "enabled by userspace" and "a display has been connected". Connector status cannot be reliably detected sometimes, in this case status will be set to unknown and it is still okay to enable it. Some drivers are also able to enable disconnected connectors.

So it is both meaningful to detect connectors in two states,

- An enabled connector, where it must have been operating in some configuration required by userspace
- A connected connector, where driver is sure a display has been connected to the port, this is the obvious case.

In the original implementation, only information of connected connectors are detected and an enabled connector without hotplug detection capability, like a Composite video connector, will be silently ignored. Switch to the new strategy to detect these connectors correctly.

References: https://github.com/torvalds/linux/blob/3efc57369a0ce8f76bf0804f7e673982384e4ac9/include/drm/drm_connector.h#L58-L89

### DIFF
--- a/src/detection/displayserver/linux/drm.c
+++ b/src/detection/displayserver/linux/drm.c
@@ -28,13 +28,19 @@ static const char* drmParseSysfs(FFDisplayServerResult* result)
         ffStrbufAppendS(&drmDir, entry->d_name);
         uint32_t drmDirWithDnameLength = drmDir.length;
 
-        ffStrbufAppendS(&drmDir, "/status");
-        char status = 'd'; // disconnected
-        ffReadFileData(drmDir.chars, sizeof(status), &status);
-        if (status != 'c') // connected
-        {
+        char buf;
+        ffStrbufAppendS(&drmDir, "/enabled");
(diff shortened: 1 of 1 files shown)


ziyao233:
btw, `neofetch` recognizes the CVBS port but does not consider its `status`/`enabled` properties.

CarterLi:
I think checking `enabled` is enough. Disabled displays (despite connected) should not be reported by fastfetch to be consistant with the behavior on Windows.

CarterLi:
By the way, doesn't the wayland or X11 detection method work for you?

CarterLi:
> I think checking `enabled` is enough. Disabled displays (despite connected) should not be reported by fastfetch. So the behavior on Linux is consistant with the behavior on Windows.

Let's change to detect enabled displays only in the next version.

ziyao233:
> By the way, doesn't the wayland or X11 detection method work for you?

There is no display server started,  composite video output is enabled directly using libdrm.

PR #1310, opened 2024-10-02
Completions (zsh, fish): check for python3

Davis-A:
Both the zsh and fish completions depend on python3. When the completitions kick in with those shells where no python3 is found it throws errors.

Check python3 is installed, otherwise return no completitions to avoid throwing errors

### DIFF
--- a/completions/fastfetch.fish
+++ b/completions/fastfetch.fish
@@ -2,6 +2,11 @@ if not type -q fastfetch
     exit
 end
 
+command -q python3
+if test $status -ne 0
+    exit
+end
+
 complete -c fastfetch -f
 
 function __fastfetch_complete_bool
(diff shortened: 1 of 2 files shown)


Davis-A:
If you're using fish (or zsh with completions turned on) and the completions kick in and you lack python3, then it throws an error.  eg

```
[17:43] adavis@carbon ~> fastfetch  fish: Unknown command: python3
/usr/share/fish/vendor_completions.d/fastfetch.fish (line 101):
' | python3 | source
    ^~~~~~^
from sourcing file /usr/share/fish/vendor_completions.d/fastfetch.fish
fastfetch                fastfetch
```

As far as i can tell no major (if any) distro packages list python3 as a dependency, so it'd be really nice to do something to avoid this.  Whether my PR is the right answer or not is in the eye of the beholder :)

CarterLi:
May I ask what os are you using?

Davis-A:
Thanks for merging.

> May I ask what os are you using?

I noticed this on a new Arch install

PR #1336, opened 2024-10-12
Linux: simplify ffNetifGetDefaultRouteImpl

apocelipes:
Use fgets to skip the first line of "/proc/net/route" since its line size is fixed.

Here's the kernel code that generate the "/proc/net/route":
image

There are 127 chars and a newline, 127+1=128.

image

<https://github.com/torvalds/linux/blob/7234e2ea0edd00bfb6bb2159e55878c19885ce68/net/ipv4/fib_trie.c#L2968>

They also say that the file format won't be changed in future, so use fgets is safe and it is simpler than a getc-loop.

### DIFF
--- a/src/common/netif/netif_linux.c
+++ b/src/common/netif/netif_linux.c
@@ -6,6 +6,8 @@
 
 #define FF_STR_INDIR(x) #x
 #define FF_STR(x) FF_STR_INDIR(x)
+// /proc/net/route has a fixed 128 chars line size, +1 for the NUL
+#define NET_ROUTE_LINE_SIZE 129
 
 bool ffNetifGetDefaultRouteImpl(char iface[IF_NAMESIZE + 1], uint32_t* ifIndex)
 {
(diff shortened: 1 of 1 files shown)


apocelipes:
I dont think using fseek on procfs/sysfs is a good idea. Actually you can find tons of problems that lseek/fseek doesn't work as expected, that why I use fgets ranther than fseek. Fortunately, it did not cause problems on the router. However if you strongly prefer to use fseek YOU CAN EXPLAIN WHY AND ASK ME TO CHANGE TO IT, OR YOU CAN ASK ME WHY NOT USE FSEEK. JUST DONT SAY NOTHING AND CHANGE IT BY YOURSELF, THAT'S NOT HOW OPEN SOURCE COMMUNITIES WORK.

apocelipes:
I don't mean to blame anyone but this has happened so many times that it has made me distrust the contribution process and feel frustrated.

CarterLi:
> Actually you can find tons of problems that lseek/fseek doesn't work as expected

Only if the files constantly changes like `/proc/meminfo`. `/proc/net/route` is not one of them.

> JUST DONT SAY NOTHING AND CHANGE IT BY YOURSELF, THAT'S NOT HOW OPEN SOURCE COMMUNITIES WORK.

That's my style. I just have no enough patience to argue something one by one.

Don't be mad, ok?

CarterLi:
> not only see an email tells me the issue/pr was closed with no other info

Still better than absolutely no email.

PR #1341, opened 2024-10-13
Processing(Linux, SunOS): simplify the code and little optimizations

apocelipes:
Use readlinkat/ffReadFileBufferRelative rather than using snprintf to make path.

We won't check the error of `open(procFilePath, ...)` since `readlinkat/ffReadFileBufferRelative` already checked it.

Optimze:
- reduce procFilePath buf size: On Linux, pid is limited up to 2^22 (in the future may be 2^30) it has at most 10 digits, kernel code is [here](https://github.com/torvalds/linux/blob/master/include/linux/threads.h#L34), 18 chars buf is suitable for the form "/proc/pid/"; on SunOS the pid could has at most 20 digits (2^64), use a 32 chars buf is enough.
- Only allocate exePath when readlinkat successed.
- Only allocate the actual size of exePath, do not allocate a PATH_MAX length memory.

Why I did the last two optimizations:

Common values ​​for PATH_MAX on Linux are 4096 or 8192, but the common exePath for example "/home/apocelipes/.local/bin/flashfetch" is only 38 bytes and "/usr/bin/bash" is only 13 bytes. If we allocate exePath with PATH_MAX size, it would waste almost 4k heap memory. This optimization waste one memcpy but the path is short in the most common cases and we no longer need to allocate a huge block of heap memory, so there's no noticeable performance regression. 

The 4k stack buf is also a big space cost, fortunately we have changed all `ffProcessGetInfoLinux`'s callers from recursion to looping early this year, so the cost is one-time and there is no worry about the stack overflow.

SunOS's PATH_MAX is commonly 1024, the waste is much smaller than on Linux but still wastes nearly 1k memory. That's why I did the same optimizations.

Memory usage before the optimization:
before

after:
after


src/common/processing_linux.c
```diff
@@ -126,10 +126,11 @@ void ffProcessGetInfoLinux(pid_t pid, FFstrbuf* processName, FFstrbuf* exe, cons
 
     #ifdef __linux__
 
-    char filePath[64];
-    snprintf(filePath, sizeof(filePath), "/proc/%d/cmdline", (int)pid);
+    char filePath[18];
+    snprintf(filePath, sizeof(filePath), "/proc/%d/", (int)pid);
+    FF_AUTO_CLOSE_FD int dfd = open(filePath, O_RDONLY | O_CLOEXEC);
```
CarterLi, line 131: You saved only one cheap `snprintf` call but wasted two expensive syscalls (`open` and `close`). It's not worth the effort.

src/common/processing_linux.c
```diff
@@ -242,24 +243,25 @@ void ffProcessGetInfoLinux(pid_t pid, FFstrbuf* processName, FFstrbuf* exe, cons
 
     #elif defined(__sun)
 
-    char filePath[PATH_MAX];
-    snprintf(filePath, sizeof(filePath), "/proc/%d/psinfo", (int) pid);
+    char filePath[32];
+    snprintf(filePath, sizeof(filePath), "/proc/%d/", (int)pid);
+    FF_AUTO_CLOSE_FD int dfd = open(filePath, O_RDONLY | O_CLOEXEC);
```
CarterLi, line 248: I don't like this

src/common/processing_linux.c
```diff
@@ -293,12 +295,14 @@ const char* ffProcessGetBasicInfoLinux(pid_t pid, FFstrbuf* name, pid_t* ppid, i
 
     #ifdef __linux__
 
-    char procFilePath[64];
+    char procFilePath[18];
+    snprintf(procFilePath, sizeof(procFilePath), "/proc/%d/", (int)pid);
+    FF_AUTO_CLOSE_FD int dfd = open(procFilePath, O_RDONLY | O_CLOEXEC);
```
CarterLi, line 300: ditto

src/common/processing_linux.c
apocelipes: All right, I'll remove this part, but the buf size will change to fit.

apocelipes: Done.

src/common/processing_linux.c
apocelipes: Done.

src/common/processing_linux.c
```diff
@@ -126,7 +126,7 @@ void ffProcessGetInfoLinux(pid_t pid, FFstrbuf* processName, FFstrbuf* exe, cons
 
     #ifdef __linux__
 
-    char filePath[64];
```
CarterLi, line 129: You won't save memory by shrinking stack variables. There is no difference as long as the stack doesn't overflow.

Stack is statically allocated when the thread is created. It's always there, no matter if you use it or not.

CarterLi: To be honest I'm not a fun of calculating *how many bytes a number string will take*. What if you make a mistake? A reasonable safe number is better for me.

src/common/processing_linux.c
```diff
@@ -139,12 +139,13 @@ void ffProcessGetInfoLinux(pid_t pid, FFstrbuf* processName, FFstrbuf* exe, cons
...
-        ssize_t length = readlink(filePath, exePath->chars, exePath->allocated - 1);
+        char buf[PATH_MAX];
+        ssize_t length = readlink(filePath, buf, PATH_MAX - 1);
         if (length > 0) // doesn't contain trailing NUL
         {
-            exePath->chars[length] = '\0';
-            exePath->length = (uint32_t) length;
+            buf[length] = '\0';
+            ffStrbufEnsureFixedLengthFree(exePath, (uint32_t)length + 1); // +1 for the NUL
+            ffStrbufAppendNS(exePath, (uint32_t)length, buf);
```
CarterLi, line 148: This does save some heap memory. Make sense.

src/common/processing_linux.c
apocelipes: `char procFilePath[PATH_MAX]` is an obvious waste. I will keep this change and discard others.

apocelipes: > Stack is statically allocated when the thread is created. It's always there, no matter if you use it or not.

The real world thing is that linux kernel only allocates a small fixed size of stack and grows it when needed (the main thread, other threads such as are created by pthread will get a fixed stack).
image

But I agree that save 100 bytes does not make sense here.

apocelipes: > To be honest I'm not a fun of calculating _how many bytes a number string will take_. What if you make a mistake? A reasonable safe number is better for me.

128 on SunOS is safe and you have already used it, it's better than PATH_MAX. I wont change other buf size.

apocelipes: Done.

CarterLi: Really? That is interesting. AFAIK a stack memory must be continuous, which means that you must move all used memory to the new memory space without modifying stack pointers (rbp rsp etc). I wonder how it can be done.

apocelipes: > Really? That is interesting. AFAIK a stack memory must be continuous, which means that you must move all used memory to the new memory space without modifying stack pointers (rbp rsp etc). I wonder how it can be done.

you can see this one: https://stackoverflow.com/questions/49595006/is-stack-memory-contiguous-physically-in-linux
Stack memory is not necessarily contiguous in the physical address space on Linux.

This one says Windows even allows discontiguous virtual memory: https://stackoverflow.com/questions/5086577/is-stack-memory-contiguous, well I only have a poor knownledge of Windows, so I cannot say this is true or not.

apocelipes: > Really? That is interesting. AFAIK a stack memory must be continuous, which means that you must move all used memory to the new memory space without modifying stack pointers (rbp rsp etc). I wonder how it can be done.

By the way, stack address, frame pointer and other things are usually virtual memory, keep it be continuous is easy because the kernel can modify the page table and make discontiguous physical memory looks like continuous. And this is no so much harm for the cache.

CarterLi: I dont think so. It is easy ONLY if you have control to the physical memory address. I dont think it is possible for a userland program.

CarterLi: If it is possible, why do we have to move memories when appending a string or array?

apocelipes: > I dont think so. It is easy ONLY if you have control to the physical memory address. I dont think it is possible for a userland program.

Stack is not controled by user space, when you add the stack pointer and access the stack memory, if it did not allocated a physical memory area, a page fault will be caused. The kernel will handle the PF and allocate physical memory, finally add it to the page table.
For detail you can see this: https://stackoverflow.com/questions/46790666/how-is-stack-memory-allocated-when-using-push-or-sub-x86-instructions

apocelipes: > If it is possible, why do we have to move memories when appending a string or array?

Why we should move these memory, because we can not control the virtual memory address on the user space. The Linux kernel only handles few special cases for us. A normal mmap used by malloc wont care about the virtual memory address. Even in the kernel only a few parts can access to physical memory and arbitrarily set the virtual memory address.
 (brk/sbrk can make the small heap allocations be continuous, but they are syscalls, not on the user land).

PR #1371, opened 2024-11-02
GPU (Apple): availability checks

khaneliman:
Getting compiler warnings for features that aren't guarded behind an availability check when targeting a lower deployment target.

Tested compiling against 10.13, 12.0, and 15.0 SDK.

### DIFF
--- a/src/detection/gpu/gpu_apple.m
+++ b/src/detection/gpu/gpu_apple.m
@@ -33,6 +33,42 @@ const char* ffGpuDetectDriverVersion(FFlist* gpus)
     return "Unsupported macOS version";
 }
 
+static void setMetalPlatformApi(id<MTLDevice> device, FFGPUResult* gpu) {
+     #ifndef MAC_OS_X_VERSION_10_15
+     if (@available(macOS 10.14, *)) {
+         if ([device supportsFeatureSet:MTLFeatureSet_macOS_GPUFamily2_v1]) {
+             ffStrbufSetStatic(&gpu->platformApi, "Metal Feature Set 2");
+             return;
+         }
+     }
+     if (@available(macOS 10.11, *)) {
+        if ([device supportsFeatureSet:MTLFeatureSet_macOS_GPUFamily1_v1]) {
+            ffStrbufSetStatic(&gpu->platformApi, "Metal Feature Set 1");
+            return;
+        }
+     }
+     #else // MAC_OS_X_VERSION_10_15
+     if (@available(macOS 13.0, *)) {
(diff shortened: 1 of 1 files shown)


src/detection/gpu/gpu_apple.m
```diff
@@ -56,8 +56,10 @@
             else if ([device supportsFeatureSet:MTLFeatureSet_macOS_GPUFamily1_v1])
                 ffStrbufSetStatic(&gpu->platformApi, "Metal Feature Set 1");
             #else // MAC_OS_X_VERSION_10_15
-            if ([device supportsFamily:MTLGPUFamilyMetal3])
-                ffStrbufSetStatic(&gpu->platformApi, "Metal 3");
+            if (@available(macOS 13.0, *)) {
```
CarterLi, line 59: You are changing the logic. On macOS 13.0+, if the device doesn't support `MTLGPUFamilyMetal3` the code won't continue to check `MTLGPUFamilyCommon3`

khaneliman: Sorry, you're right. I was trying to keep it a simple change. Looks like it didn't work with trying if logic shortcut `if (@available(macOS 13.0, *) && [device supportsFamily:MTLGPUFamilyMetal3])` so I just created a helper method for the common metal platforms and we check Metal 3 first on 13 and continue the fallback checks for others if it's not 13.

CarterLi: I really don't like the change. Since it's just a warning, does `#pragma clang diagnostic ignored "-Wxxx"` work?

khaneliman: My understanding of the availability checks are to prevent runtime crashes of this code being executed on a lower macos version. It is only a warning because the target SDK for compilation knows about the symbol.

khaneliman: I reached out to someone with more domain expertise and suggested that https://developer.apple.com/documentation/metal/mtlgpufamily/mtlgpufamilymetal3 the symbol being a constant might not be a runtime crash, just an ugly compiler warning that could actually be fine... But, it looks like we use a `#ifndef MAC_OS_X_VERSION_10_15` wrapping calls that contain macos 13 features which isn't correct and we might be able to just bump that version check.

CarterLi: > the symbol being a constant might not be a runtime crash

Definitely not. You can test it with `[device supportsFamily:(MTLGPUFamily)999999]`

CarterLi:
I still don't like the change. Let's just leave as it was.

PR #1382, opened 2024-11-09
IO: optimize createSubfolders

apocelipes:
File names have known max length limits, so we can use a fixed length stack buf rather than heap allocations.

This reduces 2 heap allocations on the path "/home/apocelipes/.config/fastfetch" and makes `createSubfolders` run 8% faster:

benchmark

After this optimization, the paramater `fileName`'s length must be less than the PATH_MAX/MAX_PATH, and `open`/`CreateFileA` has already checked this for us, so it's not worth to check it again in the `createSubfolders`.

### DIFF
--- a/src/common/io/io_unix.c
+++ b/src/common/io/io_unix.c
@@ -23,13 +23,14 @@
 
 static void createSubfolders(const char* fileName)
 {
-    FF_STRBUF_AUTO_DESTROY path = ffStrbufCreate();
-
+    char path[PATH_MAX];
     char *token = NULL;
+    char *pathTail = path;
     while((token = strchr(fileName, '/')) != NULL)
     {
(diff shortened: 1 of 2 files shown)


apocelipes:
It's not worth using strncat, it is noticeably slower than pathTail in benchmarks. Let the code be a bit longer can get better performance.

CarterLi:
I don't like the change. Maybe you care the performance of functions that only run once, I don't.

apocelipes:
> only run once.

Currently, but may be reused in the future.

Also, the new code is simple, we can do it better. fastfetch runs fastly is not only because it's wrote by c.

apocelipes:
I reused ffStrCopyN. Simpler code and better performance.

PR #1392, opened 2024-11-12
Python script to generate man page

### LINKED ISSUE #830: [FEAT] More documentation in the Manpage

Tiuxi:
Added the script `scripts/gen-man.py` that generate a man page using the `src/data/help.json` file, and print the generated file to stdout.

close #830


CarterLi:
```
scripts/gen-man.py:146: SyntaxWarning: invalid escape sequence '\-'
  print(f"\-{ option['short'] }", end="")
scripts/gen-man.py:153: SyntaxWarning: invalid escape sequence '\-'
  print(f"\-\-{ option['long'] }", end="")
```

CarterLi:
I'd like to keep the `Config Files` section in the old man file.

CarterLi:
You can acquire the fastfetch version from `CMakeList.txt` using RegExp `^    VERSION (\d+\.\d+\.\d+)$`

Tiuxi:
Is it relevant to also put a short example of a config file (E.g from a preset) at the end ?

CarterLi:
Yes. You should also metion the Github Wiki page.

PR #1403, opened 2024-11-18
Fixed the destination where fastfetch.1 is generated.

phoepsilonix:
`cmake -B build -S fastfetch` generates fastfetch.1 in the directory where cmake is executed.
This patch is a fix to generate fastfetch.1 in PROJECT_BINARY_DIR, that is, the build directory.


CarterLi:
Thanks. A hotfix release will be cut to fix this issue.

Issue #1413, opened 2024-11-24
[FEAT] dual CPU support


AzimovParviz:
Be sure to read [FAQ](https://github.com/fastfetch-cli/fastfetch?tab=readme-ov-file#faq) before submitting a new issue.

## Wanted features:

I want fastfetch to differentiate if there's multiple physical CPUs on the system. 

## Motivation:

I have no hard numbers to back me up, but I assume there's plenty of people using fastfetch on their dual Xeon workstations running Windows/macOS/Linux.

I have came up with the following macOS only implementation https://github.com/AzimovParviz/fastfetch/commit/d3e8416f73f55ea40bd7dc53934d6b554f11a895 in my fork.

This is how it looks on my MacPro5,1 with dual quad core Xeons 
![fastfetch dual cpu example](https://github.com/user-attachments/assets/9cd96e2a-0955-4bf6-85cc-a48025b36374)

I also did https://github.com/AzimovParviz/fastfetch/commit/3355cd505d6c2f2a2b85aef4d0cf8f3dd807a928 to account for the "2 x" string to the output more clear, but I am not confident in whether this is the correct way to go about it

I did not find any contribution guidelines outside of line 206 of README.md so I am doing a feature request before doing a PR

CarterLi:
Well, fastfetch shows 16 threads. Seems that it already takes dual CPU into account.

Another question: What if user installs 2 different CPU in 1 machine. Is it possible?

AzimovParviz:
> Well, fastfetch shows 16 threads. Seems that it already takes dual CPU into account.
Yeah but wouldn't it be better if outright stated the number of CPUs like on the screenshot? 

> Another question: What if user installs 2 different CPU in 1 machine. Is it possible?
I definitely know that you need to have a matching pair on MacOS. I cannot say for sure for other systems. Cursory google search tells that most motherboards won't even post but some systems seem to work when mixing CPUs. But I never seen it and can't seem to find a video of someone doing it

CarterLi:
If you can figger out how to detect it in Linux and Windows, I am happy to have it supported in fastfetch.

AzimovParviz:
> If you can figger out how to detect the CPU count in Linux and Windows, I am happy to have it supported in fastfetch.

I'll take a shot at it. I never developed on Windows, I spent the entire evening yesterday trying to just build the project. It fails to find vulkan headers whenever I try building. It also won't build unless I do `cmake -G Ninja`

Here's how it looks like
[...]

PR #1417, opened 2024-11-25
Support for dual CPU systems 

AzimovParviz:
https://github.com/fastfetch-cli/fastfetch/issues/1413

macOS and Windows support done

Here is how it looks on macOS and Windows for now:

389268485-9cd96e2a-0955-4bf6-85cc-a48025b36374

fastfetch dual CPU windows


src/modules/cpu/cpu.c
```diff
@@ -68,7 +71,12 @@ void ffPrintCPU(FFCPUOptions* options)
             if(coreTypes.length > 0)
                 ffStrbufAppendF(&str, " (%s)", coreTypes.chars);
             else if(cpu.coresOnline > 1)
-                ffStrbufAppendF(&str, " (%u)", cpu.coresOnline);
+            {
+                if(cpu.cpuCount > 1)
+                    ffStrbufAppendF(&str, " (%u)", cpu.coresOnline / 2);
```
CarterLi, line 76: `cpu.coresOnline / cpu.cpuCount`?

What if you disable some cores in BIOS?

AzimovParviz: Good question. Things like these is why I was not 100% sure if it's better to do it like this or just leave as it is now (total sum between the processor packages). The alternative could be doing a line for each CPU (like it would be done for laptops with iGPU and dGPU) and then listing each CPU's coresOnline?

CarterLi: I prefer leaving as it is now

CarterLi: Another question: is an [AMD CCD](https://www.reddit.com/r/Amd/comments/io3twa/what_is_ccdcore_complex_die/) considered a package?

xiaoran007: > Another question: is an [AMD CCD](https://www.reddit.com/r/Amd/comments/io3twa/what_is_ccdcore_complex_die/) considered a package?

No, on Linux, the entire CPU is considered a "Physical CPU", rather than a CCD being considered a "Physical CPU". I confirm that all cores of AMD EPYC 9754 CPU (128 cores) are marked with id 0. But I'm not sure about the implementation in windows because I don't have a corresponding windows device.

src/detection/cpu/cpu.h
```diff
@@ -15,6 +15,7 @@ typedef struct FFCPUResult
     FFstrbuf name;
     FFstrbuf vendor;
 
+    uint16_t cpuCount;
```
AzimovParviz, line 18: Should this be renamed to something else like packageCount ?

src/detection/cpu/cpu_linux.c
```diff
@@ -247,6 +248,9 @@ static const char* parseCpuInfo(
             #if !(__arm__ || __aarch64__)
             (cpu->name.length == 0 && ffParsePropLine(line, "model name :", &cpu->name)) ||
             (cpu->vendor.length == 0 && ffParsePropLine(line, "vendor_id :", &cpu->vendor)) ||
+            //Is it cheaper to just parse every physical id or to check if it's already set to the parsed value?
```
AzimovParviz, line 251: I am not sure whether it's the most optimal way to go about things. The intended logic is to see how many different physical ids are present in cpuinfo.

src/detection/cpu/cpu.h
CarterLi: Not sure if `physical id` in Linux is the same as package in Windows and macOS. I guess cpuCount is ok.

src/detection/cpu/cpu_linux.c
CarterLi: I guess we just need to fix the maximum id?

src/detection/cpu/cpu_linux.c
```diff
@@ -247,6 +248,9 @@ static const char* parseCpuInfo(
...
+            (cpuPhysicalId->length == 0 && ffParsePropLine(line, "physical id:", cpuPhysicalId)) ||
+            (cpuPhysicalId->length > 0 && ffParsePropLine(line, "physical id:", cpuPhysicalId)) ||
```
CarterLi, line 253: This is weird. `cpuPhysicalId->length == 0 || cpuPhysicalId->length > 0` doesn't make sense to me.

src/detection/cpu/cpu_linux.c
CarterLi: I don't think this code works. Currently [the loop exits after the first core entry](https://github.com/fastfetch-cli/fastfetch/pull/1417/files#diff-6f3a2f89998ef2269032903b3586e2c86b1f0720760190e4785e7db3a0c6a72fR237-R243), while `physical id` requires enumerating every entry.

CarterLi:
I will do the rest change

PR #1429, opened 2024-12-08
CMake(Linux): make "-fno-omit-frame-pointer" configurable in release …

apocelipes:
…mode.

Starting last year, Fedora and Ubuntu began to require softwares in the repository to turn on the "-fno-omit-frame-pointer" and "-mno-omit-leaf-frame-pointer" compiler options by default. This year Arch Linux also began to make similar requirements.

Fedora: <https://fedoraproject.org/wiki/Changes/fno-omit-frame-pointer>
Ubuntu: <https://www.phoronix.com/news/Ubuntu-Frame-Pointers-Default>
Arch Linux: <https://rfc.archlinux.page/0026-fno-omit-frame-pointer/>

So add an option and set its default value to OFF, packagers can set to ON if they need it. This might make those packagers' lives easier.

Currently the frame pointer is only valuable under Linux so it is only enabled when the build target is Linux.

This option is also not enabled in Debug mode, for two reasons:
1. Debug mode already sets "-fno-omit-frame-pointer", that's enough.
2. Frame pointers are not only for debugging but for tools like perf and eBPF to do profiling.

Test:

Turn on the option:
on

Default/Turn off:
off

image

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -91,6 +91,7 @@ option(SET_TWEAK "Add tweak to project version" ON) # This is set to off by gith
 option(IS_MUSL "Build with musl libc" OFF) # Used by Github Actions
 option(INSTALL_LICENSE "Install license into /usr/share/licenses" ON)
 option(ENABLE_EMBEDDED_PCIIDS "Embed pci.ids into fastfetch, requires `python`" OFF)
+option(ENABLE_FRAME_POINTERS "Build fastfetch with frame pointers" OFF) # For Fedora/Ubuntu/Arch Linux packagers
 
 set(BINARY_LINK_TYPE_OPTIONS dlopen dynamic static)
 set(BINARY_LINK_TYPE dlopen CACHE STRING "How to link fastfetch")
@@ -137,6 +138,10 @@ set(WARNING_FLAGS "-Wall -Wextra -Wconversion -Werror=uninitialized -Werror=retu
 
 set(CMAKE_C_STANDARD 11)
 set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${WARNING_FLAGS} -Werror=incompatible-pointer-types -Werror=implicit-function-declaration -Werror=int-conversion")
+
+if(LINUX AND ENABLE_FRAME_POINTERS AND NOT CMAKE_BUILD_TYPE STREQUAL "Debug")
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer")
+endif()
 
 if(WIN32 OR ENABLE_DIRECTX_HEADERS)
     enable_language(CXX)


CarterLi:
Why not `CFLAGS="-fno-omit-frame-pointer -mno-omit-leaf-frame-pointer" cmake ...`

apocelipes:
> Why not `CFLAGS="-fno-omit-frame-pointer -mno-omit-leaf-frame-pointer" cmake ...`

It's too long and the packager should check the whole cmakelists file to ensure the flags won't be overrode, maybe too difficult to some one (a packager not need to be a programmer).

Sometimes I also need this for profiling, so I made my life easier too.

CarterLi:
> > Why not `CFLAGS="-fno-omit-frame-pointer -mno-omit-leaf-frame-pointer" cmake ...`
> 
> It's too long and the packager should check the whole cmakelists file to ensure the flags won't be overrode, maybe too difficult to some one (a packager not need to be a programmer).

There's `ninja --verbose`. There are thousands of packages but I don't think they can require every package to add this kind of option.

> 
> Sometimes I also need this for profiling, so I made my life easier too. To be honest, I can't remember options like "-mno-omit-leaf-xxxx", but I can barely remember ENABLE_FRAME_POINTERS.

What if you want `-mno-omit-leaf-yyyy` later? Fastfetch needs to add a new option again to fulfil your personal requirements?

No.

apocelipes:
That's not my PERSONAL requirement. THAT'S FEDORA/UBUNTU/ARCH's REQUIREMENTS.

apocelipes:
> There's ninja --verbose. There are thousands of packages but I don't think they can require every package to add this kind of option.

No, they can.

apocelipes:
OK, seems like u never read links that I posted and only want to against me. It's wasting time.

CarterLi:
Have you ever used CFLAGS??? `CFLAGS` is the [standard environment variable](https://en.wikipedia.org/wiki/CFLAGS) that can be set globally. Both [fedora](https://src.fedoraproject.org/rpms/fastfetch/blob/rawhide/f/fastfetch.spec#_106) and [debian](https://salsa.debian.org/debian/fastfetch/-/blob/debian/latest/debian/rules?ref_type=heads#L9) use helper scripts to invoke cmake. If I am the package manager, I will just set `CFLAGS="-fno-omit-frame-pointer -mno-omit-leaf-frame-pointer"` in the helper script, and all packages that use the script will have the flags set.

> That's not my PERSONAL requirement. THAT'S FEDORA/UBUNTU/ARCH's REQUIREMENTS.

> Sometimes I also need this for profiling, so I made my life easier too. To be honest, I can't remember options like "-mno-omit-leaf-xxxx", but I can barely remember ENABLE_FRAME_POINTERS.

This is your requirements. As for now, no mantainers of fastfetch distro packages made a request of adding this kind of **cmake option**. ENABLE_FRAME_POINTERS is your idea and named by you. I don't think you discussed it with any real fastfetch package mantainers.

CarterLi:
> OK, seems like u never read links that I posted and only want to against me. It's wasting time.

Again, please make requests that really solve problems. Otherwise, it's wasting time. Yes.

PR #1441, opened 2024-12-15
improved debian ascii logo

L1kwidus:
improved/fixed debian ascii logo


CarterLi:
Didn't see many differences though

PR #1447, opened 2024-12-18
Refactor Armbian version detection and add support for ubuntu-based armbian detection

JohnTheCoolingFan:
Ubuntu can be used to build Armbian, it is one of the official bases for Armbian. This allows the detection of Armbian in case it is based on Ubuntu.

### DIFF
--- a/src/detection/os/os_linux.c
+++ b/src/detection/os/os_linux.c
@@ -49,10 +49,7 @@ static bool parseOsRelease(const char* fileName, FFOSResult* result)
 // Get Armbian version properties and set idLike based on the Armbian image basis
 FF_MAYBE_UNUSED static void getArmbianVersion(FFOSResult* result)
 {
-    if(ffStrbufIgnCaseEqualS(&result->id, "ubuntu"))
-        ffStrbufSetS(&result->idLike, "ubuntu");
-    else if(ffStrbufIgnCaseEqualS(&result->id, "debian"))
-        ffStrbufSetS(&result->idLike, "debian");
+    ffStrbufSet(&result->idLike, &result->id);
     ffStrbufSetS(&result->id, "armbian");
     ffStrbufClear(&result->versionID);
     uint32_t versionStart = ffStrbufFirstIndexC(&result->prettyName, ' ') + 1;
--- a/src/detection/os/os_linux.c
+++ b/src/detection/os/os_linux.c
@@ -46,34 +46,46 @@ static bool parseOsRelease(const char* fileName, FFOSResult* result)
     });
 }
 
+// Get Armbian version properties and set idLike based on the Armbian image basis
+FF_MAYBE_UNUSED static void getArmbianVersion(FFOSResult* result)
+{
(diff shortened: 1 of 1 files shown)


CarterLi:
Please reuse code of debian flavors.

JohnTheCoolingFan:
The debian-derived detection function doesn't run if the id is Ubuntu, which is the case for an ubuntu-based Armbian image. It will still need to be handled in both Ubuntu and debian variant detection. If you want me to just refactor so that Armbian detection uses a common function in both Ubuntu and debian flavours detection then that's fine, but I don't completely understand what you mean.

CarterLi:
> refactor so that Armbian detection uses a common function in both Ubuntu and debian flavours detection

I mean this

JohnTheCoolingFan:
Alright, refactored the armbian detection. The version detection is in a single function, which also sets `idLike` based on the original `id`. This is then used in both `getUbuntuFlavour` and `detectDebianDerived`.

src/detection/os/os_linux.c
```diff
@@ -46,13 +46,47 @@ static bool parseOsRelease(const char* fileName, FFOSResult* result)
     });
 }
 
+// Get Armbian version properties and set idLike based on the Armbian image basis
+FF_MAYBE_UNUSED static void getArmbianVersion(FFOSResult* result)
+{
+    if(ffStrbufIgnCaseEqualS(&result->id, "ubuntu"))
+        ffStrbufSetS(&result->idLike, "ubuntu");
+    else if(ffStrbufIgnCaseEqualS(&result->id, "debian"))
+        ffStrbufSetS(&result->idLike, "debian");
```
CarterLi, line 55: ffStrbufSet(&result->idLike, &result->id);

src/detection/os/os_linux.c
```diff
@@ -46,13 +46,47 @@ static bool parseOsRelease(const char* fileName, FFOSResult* result)
...
+        getArmbianVersion(result);
+        return true;
+    }
+    else if (ffStrbufStartsWithS(&result->prettyName, "Armbian-unofficial ")) // Unofficial Armbian image built from source
+    {
+        ffStrbufSetS(&result->name, "Armbian (custom build)");
+        getArmbianVersion(result);
+        return true;
+    }
+    return false;
```
CarterLi, line 78:
```c
if (ffStrbufStartsWithS(&result->prettyName, "Armbian "))
    ffStrbufSetS(&result->name, "Armbian");
else if (ffStrbufStartsWithS(&result->prettyName, "Armbian-unofficial ")
    ffStrbufSetS(&result->name, "Armbian (custom build)");
else
    return false;

// Other code in getArmbianVersion
```

CarterLi:
Thanks.

CarterLi:
Since it's called **arm**bian, I assume it supports only arm32 and aarch64? Is `#if __arm__ || __aarch64__` ok?

JohnTheCoolingFan:
While the project was initially aimed at arm SBCs, it currently supports x86_64 as well, and I think I've heard something about risc-v boards. So no, it shouldn't be limited to arm.

PR #1460, opened 2024-12-22
[LOGO] Update logo of NurOS

GoldenVadim:


CarterLi:
Your logo is too large to be acceptable.

GoldenVadim:
> Your logo is too large to be acceptable.

Which image size I must have to convert it to ASCII art?

PR #1481, opened 2025-01-05
Add the Lubuntu logo

tsimonq2:
This is the same logo we contributed to Neofetch a while back.


tsimonq2:
Also, Lubuntu uses LXQt, not LXDE. I've corrected that as well.

tsimonq2:
image

Looks good to me!

CarterLi:
Thanks

PR #1485, opened 2025-01-06
feat: add snigdhaos ascii

eshanized:


CarterLi:
You should also update `src\logo\builtin.c`

PR #1536, opened 2025-02-03
Add missing new lines after printing, like neofetch

sirenuf:
Adds the missing new lines after printing the colour scheme showcase:

Before:
before2

After:
after2

Maybe this needs to edited, as I am not sure if the new lines are in an appropriate location (just a printf call in the main function, inside fastfetch.c)


CarterLi:
Dup #989 & #990

PR #1538, opened 2025-02-04
Minimal Haiku implementation

mmuman:
This is a first try at supporting Haiku, with the minimum to compile.

Copie d'écran_20250204_114510

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,6 +30,8 @@ elseif("${CMAKE_SYSTEM_NAME}" STREQUAL "DragonFly")
     set(DragonFly TRUE CACHE BOOL "..." FORCE)
 elseif("${CMAKE_SYSTEM_NAME}" STREQUAL "SunOS")
     set(SunOS TRUE CACHE BOOL "..." FORCE)
+elseif("${CMAKE_SYSTEM_NAME}" STREQUAL "Haiku")
+    set(Haiku TRUE CACHE BOOL "..." FORCE)
 elseif(NOT APPLE AND NOT WIN32)
     message(FATAL_ERROR "Unsupported platform: ${CMAKE_SYSTEM_NAME}")
 endif()
@@ -139,7 +141,7 @@ set(WARNING_FLAGS "-Wall -Wextra -Wconversion -Werror=uninitialized -Werror=retu
 set(CMAKE_C_STANDARD 11)
 set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${WARNING_FLAGS} -Werror=incompatible-pointer-types -Werror=implicit-function-declaration -Werror=int-conversion")
 
-if(WIN32 OR ENABLE_DIRECTX_HEADERS)
+if(WIN32 OR HAIKU OR ENABLE_DIRECTX_HEADERS)
     enable_language(CXX)
     set(CMAKE_CXX_STANDARD 17)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${WARNING_FLAGS}")
@@ -1088,6 +1090,74 @@ elseif(SunOS)
         src/util/platform/FFPlatform_unix.c
         src/util/binary_linux.c
     )
+elseif(Haiku)
+    list(APPEND LIBFASTFETCH_SRC
+        src/common/io/io_unix.c
+        src/common/netif/netif_haiku.c
+        src/common/networking_linux.c
+        src/common/processing_linux.c
+        src/detection/battery/battery_nosupport.c
+        src/detection/bios/bios_nosupport.c
+        src/detection/board/board_nosupport.c
+        src/detection/bootmgr/bootmgr_nosupport.c
+        src/detection/brightness/brightness_nosupport.c
+        src/detection/btrfs/btrfs_nosupport.c
+        src/detection/chassis/chassis_nosupport.c
+        src/detection/cpu/cpu_nosupport.c
+        src/detection/cpucache/cpucache_nosupport.c
+        src/detection/cpuusage/cpuusage_nosupport.c
+        src/detection/cursor/cursor_nosupport.c
+        src/detection/bluetooth/bluetooth_nosupport.c
+        src/detection/bluetoothradio/bluetoothradio_nosupport.c
+        src/detection/disk/disk_nosupport.c
+        src/detection/dns/dns_linux.c
+        src/detection/physicaldisk/physicaldisk_nosupport.c
+        src/detection/physicalmemory/physicalmemory_nosupport.c
+        src/detection/diskio/diskio_nosupport.c
+        src/detection/displayserver/displayserver_haiku.cpp
+        src/detection/font/font_haiku.cpp
+        src/detection/gpu/gpu_nosupport.c
+        src/detection/gpu/gpu_pci.c
+        src/detection/gtk_qt/gtk.c
+        src/detection/host/host_nosupport.c
+        src/detection/icons/icons_nosupport.c
+        src/detection/initsystem/initsystem_nosupport.c
+        src/detection/keyboard/keyboard_nosupport.c
+        src/detection/libc/libc_nosupport.c
+        src/detection/lm/lm_nosupport.c
+        src/detection/loadavg/loadavg_nosupport.c
+        src/detection/locale/locale_linux.c
+        src/detection/localip/localip_linux.c
+        src/detection/gamepad/gamepad_nosupport.c
+        src/detection/media/media_nosupport.c
+        src/detection/memory/memory_haiku.c
+        src/detection/mouse/mouse_nosupport.c
+        src/detection/netio/netio_nosupport.c
+        src/detection/opengl/opengl_linux.c
+        src/detection/os/os_uname.c
+        src/detection/packages/packages_haiku.c
+        src/detection/poweradapter/poweradapter_nosupport.c
+        src/detection/processes/processes_haiku.c
+        src/detection/gtk_qt/qt.c
+        src/detection/sound/sound_nosupport.c
+        src/detection/swap/swap_haiku.c
+        src/detection/terminalfont/terminalfont_linux.c
+        src/detection/terminalshell/terminalshell_linux.c
+        src/detection/terminalsize/terminalsize_linux.c
+        src/detection/theme/theme_nosupport.c
+        src/detection/tpm/tpm_nosupport.c
+        src/detection/uptime/uptime_haiku.c
+        src/detection/users/users_linux.c
+        src/detection/wallpaper/wallpaper_nosupport.c
+        src/detection/wifi/wifi_nosupport.c
+        src/detection/wm/wm_nosupport.c
+        src/detection/de/de_nosupport.c
+        src/detection/wmtheme/wmtheme_nosupport.c
+        src/detection/camera/camera_nosupport.c
+        src/detection/zpool/zpool_nosupport.c
+        src/util/platform/FFPlatform_unix.c
+        src/util/binary_linux.c
+    )
 endif()
 
 if(ENABLE_DIRECTX_HEADERS)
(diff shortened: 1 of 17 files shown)


CMakeLists.txt
```diff
@@ -1088,6 +1090,74 @@ elseif(SunOS)
         src/util/platform/FFPlatform_unix.c
         src/util/binary_linux.c
     )
+elseif(Haiku)
```
CarterLi, line 1093: Does Haiku support OpenGL, pulseaudio or something?

src/common/netif/netif_haiku.c
```diff
@@ -0,0 +1,64 @@
...
+    struct ifconf config;
+    config.ifc_len = sizeof(config.ifc_value);
+    if (ioctl(pfRoute, SIOCGRTSIZE, &config, sizeof(struct ifconf)) < 0)
+        return false;
+
+    int size = (size_t)config.ifc_value;
+    if (size == 0)
+        return false;
+
+    void *buffer = malloc(size);
```
CarterLi, line 29: FF_AUTO_FREE

src/common/netif/netif_haiku.c
```diff
@@ -0,0 +1,64 @@
...
+    if (buffer == NULL) {
+        return false;
+    }
+
+    config.ifc_len = size;
+    config.ifc_buf = buffer;
+    if (ioctl(pfRoute, SIOCGRTTABLE, &config, sizeof(struct ifconf)) < 0)
+        return false;
```
CarterLi, line 37: Variable `buffer` leaks here

src/detection/displayserver/displayserver_haiku.cpp
```diff
@@ -0,0 +1,73 @@
...
+#include <Application.h>
+#include <Screen.h>
+
+extern "C" void ffConnectDisplayServerImpl(FFDisplayServerResult* ds);
+
+static void detectDisplays(FFDisplayServerResult* ds)
+{
+    // We need a valid be_app to query the app_server here.
+    BApplication app("application/x-vnd.fastfetch-cli-fastfetch");
+    BScreen s; // default screen is the main one
```
CarterLi, line 18: `BScreen s{};` or it looks uninitialized to me in a mainly-C project.

src/detection/displayserver/displayserver_haiku.cpp
```diff
@@ -0,0 +1,73 @@
...
+            continue;
+
+        display_mode mode;
+        if (s.GetMode(&mode) != B_OK)
+            continue;
+
+        FF_STRBUF_AUTO_DESTROY name = ffStrbufCreateA(128);
+        monitor_info monitor;
+        // WARNING: This is experimental new Haiku API
+        err = s.GetMonitorInfo(&monitor);
```
CarterLi, line 34: status_t err =

src/detection/localip/localip_linux.c
```diff
@@ -128,8 +132,12 @@ const char* ffDetectLocalIps(const FFLocalIpOptions* options, FFlist* results)
 
     for (struct ifaddrs* ifa = ifAddrStruct; ifa; ifa = ifa->ifa_next)
     {
-        if (!ifa->ifa_addr || !(ifa->ifa_flags & IFF_RUNNING))
+        if (!ifa->ifa_addr)
             continue;
+#ifdef IFF_RUNNING
+        if (!(ifa->ifa_flags & IFF_RUNNING))
```
CarterLi, line 138: How can we know if it's running then?

CarterLi:
Thanks for your interest. Can we set up a CI to ensure it won't be broken unexpectedly?

CMakeLists.txt
mmuman: OpenGL yes, we have our own audio framework (MediaKit)… I wanted to have something that compiles before expanding to these.

src/common/netif/netif_haiku.c
mmuman: Yeah I suspected there was something like that to use, thanks.

src/detection/displayserver/displayserver_haiku.cpp
mmuman: It's an object, in C++ the default ctor is called, but OK.

src/detection/localip/localip_linux.c
mmuman: AFAIK this is not really used in most drivers on Linux anyway, it's here for BSD compatibility.

I'll add our own extra flags that I missed (IFF_LINK,…CONFIGURING…).

mmuman:
I'm not sure we can get it into the GitHub CI stuff, since we build the cross-compiler from our own build system only to generate the image…
We do have our own buildbots but they generate the official packages.

mmuman:
> Thanks for your interest. Can we set up a CI to ensure it won't be broken unexpectedly?

So I asked around, we have:

- https://github.com/korli/haiku-builder is used by  various haikuarchives projects, for example https://github.com/haikuarchives/weather
- there is also a docker image that some projects like genio use: https://github.com/Genio-The-Haiku-IDE/Genio/blob/main/.github/workflows/build.yml

src/util/platform/FFPlatform_unix.c
```diff
@@ -47,6 +47,8 @@ static void getExePath(FFPlatform* platform)
         ssize_t exePathLen = readlink("/proc/self/path/a.out", exePath, sizeof(exePath) - 1);
         if (exePathLen >= 0)
             exePath[exePathLen] = '\0';
+    #elif defined(__HAIKU__)
+        size_t exePathLen = 0; // TODO
```
jmairboeck, line 51: This can be implemented using `get_next_image_info` and searching for the `B_APP_IMAGE` (see https://www.haiku-os.org/legacy-docs/bebook/TheKernelKit_Images.html#get_next_image_info)

mmuman: Yes that's the intent.

mmuman: Actually there's a private get_app_path call that does that.

mmuman: Done.

mmuman:
- fixed boot time
- patch "BePC" as "i386" architecture
- add proper config dir.

mmuman:
- fixed OS in version info
- we have a non-standard path for `/etc/resolv.conf`

mmuman:
- added font support

mmuman:
I think I picked all the low-hanging fruits, maybe we can get this in already?

mmuman:
For the record, the JSON output: [ff.json](https://github.com/user-attachments/files/18664652/ff.json)

CarterLi:
The `IFF_` string prefix in localip_linux.c should be removed. I'm ok with others.

mmuman:
Oh right, missed that!

CarterLi:
Thanks for your contribution. I wonder how the output is comparing to neofetch's

mmuman:
Much fancier:
Copie d'écran_20250205_021115

CarterLi:
https://github.com/fastfetch-cli/fastfetch/blob/01f400c15f1bff8217c9d4c6f24316b7b254757d/CMakeLists.txt#L1137

This file doesn't exist. Does your code really compile?

Begasus:
> https://github.com/fastfetch-cli/fastfetch/blob/01f400c15f1bff8217c9d4c6f24316b7b254757d/CMakeLists.txt#L1137
> 
> This file doesn't exist. Does your code really compile?

That should be `os_haiku.c`, I think @mmuman used the nightly version for this, on R1B5 64bit I still need to apply a seperate patch for `pipe2` and `memmem`.

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 1a8b39d..9435569 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1607,6 +1607,7 @@ elseif(Haiku)
     target_link_libraries(libfastfetch
         PRIVATE "network"
         PRIVATE "be"
+        PRIVATE "gnu"
     )
 endif()
 
diff --git a/src/common/processing_linux.c b/src/common/processing_linux.c
index 63d3951..02d3bf8 100644
--- a/src/common/processing_linux.c
+++ b/src/common/processing_linux.c
@@ -34,7 +34,7 @@ enum { FF_PIPE_BUFSIZ = 8192 };
 
 static inline int ffPipe2(int *fds, int flags)
 {
-    #ifdef __APPLE__
+    #if defined(__APPLE__) || defined(__HAIKU__)
         if(pipe(fds) == -1)
             return -1;
         fcntl(fds[0], F_SETFL, fcntl(fds[0], F_GETFL) | flags);
-- 
```

fastfetch

CarterLi:
@Begasus Please test the newest commit

Some issue I found from your screenshot

1. Seems the monitor name isn't available on current version
2. The terminal name is wrong. I think ffProcessGetInfoLinux and ffProcessGetBasicInfoLinux in `processing_linux.c` lack a proper implementation for Haiku
3. The ifa name in Haiku is a device path?
4. fastfetch should report `R1B5` instead of `1`

Begasus:
Thanks so far! Looks promising (mind you I'm no developer, @mmuman is the real one :)

No patching with master checkout:

fastfetch_2

mmuman:
> https://github.com/fastfetch-cli/fastfetch/blob/01f400c15f1bff8217c9d4c6f24316b7b254757d/CMakeLists.txt#L1137
> 
> This file doesn't exist. Does your code really compile?

Oops, thought I replaced it with the haiku one, that was a test.

mmuman:
>     1. Seems the monitor name isn't available on current version

The code is there but the API returns an error, so either it's not implemented at all, or at least on VESA mode.

>     2. The terminal name is wrong. I think ffProcessGetInfoLinux and ffProcessGetBasicInfoLinux in `processing_linux.c` lack a proper implementation for Haiku

That can probably be fixed.

>     3. The ifa name in Haiku is a device path?

For most network interfaces yes, for historical reasons.

>     4. fastfetch should report `R1B5` instead of `1`

Yes, I didn't find an easy way to find that yet, we'd have to dig the resources in the kernel file or something.

CarterLi:
`--bright-color false` may help. If the terminal reports `$COLORFGBG`, fastfetch will detect whether the terminal is in light mode and set `--bright-color false` by default

Alternatively, we may set https://github.com/fastfetch-cli/fastfetch/blob/8774fb3fb7c9622c3c0b88f2d02dbc601fb09df9/src/logo/builtin.c#L2011 to `FF_COLOR_FG_DEFAULT`, so that fastfetch will use black color in light mode.

src/detection/packages/packages_haiku.c
```diff
@@ -0,0 +1,13 @@
...
+#include "common/io/io.h"
+
+void ffDetectPackagesImpl(FFPackagesResult* result, FFPackagesOptions* options)
+{
+    // TODO: Use the Package Kit C++ API instead (would account for disabled packages)
+
+    if (!(options->disabled & FF_PACKAGES_FLAG_PKG_BIT))
+        result->pkg = ffPackagesGetNumElements(FASTFETCH_TARGET_DIR_ROOT "/system/packages", false);
+    if (!(options->disabled & FF_PACKAGES_FLAG_PKGSRC_BIT))
+        result->pkgsrc = ffPackagesGetNumElements(FASTFETCH_TARGET_DIR_ROOT "/boot/home/config/packages", false);
```
jmairboeck, line 12: Usage of `pkg` and `pkgsrc` here doesn't really make sense. These are the same kind of packages (Haiku packages, "hpkg"), but rather system packages and user packages, respectively.

I see that there is such a distinction already for other packaging systems (e.g. for flatpak [here](https://github.com/fastfetch-cli/fastfetch/blob/f77a9c185df400eaaceedcad4294c75b5853133d/src/detection/packages/packages.h#L16-L17)). So I think it would make sense to add `haikuSystem` and `haikuUser` there as well.

mmuman: Yes, I just used that because I wasn't sure how it was categorized, but that'd be better.
Ideally we'd query the Package Kit because some packages could be disabled.

mmuman:
Hmm the newest commits fail to build on 32bit…

Begasus:
https://github.com/fastfetch-cli/fastfetch/commit/437006f5f104767cceaae6db8c67b6d228e88a4f still builds here on R1B5 32bit.

PR #1546, opened 2025-02-08
refactor(json): use default config file if the user config is empty

nnyyxxxx:
if the user config is empty we get a failed to parse json error, instead of erroring out we should just show the default config since there is technically nothing wrong with the users config file if its empty :)

minor variable naming change as well for readability

potentially fixes a case where we could read twice instead of once

removes a redundant || false condition (habbit probs)


CarterLi:
I don't like the change. You read config files twice unnecessarily. Meanwhile, I don't like your habbit.

> there is technically nothing wrong with the users config file if its empty

An empty string is not a valid JSON string, same as `"" != NULL`. Just remove the file.

CarterLi:
I don't like the change. You read config files twice unnecessarily. Meanwhile, I don't like your habbit.

> there is technically nothing wrong with the users config file if its empty

An empty string is not a valid JSON string, same as `"" != NULL`. Just remove the file.

PR #1547, opened 2025-02-08
Combine all armbian variants

JohnTheCoolingFan:
Change armbian detection to detect by checking `PRETTY_NAME` if it starts with `"Armbian"`, without a space on the end. This detects additional 2 values that can be there: community releases and security (kali) images.
The detection was reported to be broken for community support images, and should work for other possible Armbian image variants in the future. Wrote a list of strings that the pretty name can start with right now, in a code comment.


CarterLi:
Thanks.

PR #1557, opened 2025-02-11
[chore] Update README.md

Drsheppard01:
move repology badge to install section
customize with `columns` and `minversion` tags


CarterLi:
Why?

Drsheppard01:
It will be make more visable
In left column users can check install commands, in right column users can check which repo has it

PR #1564, opened 2025-02-15
Haiku fixes

mmuman:
Fix 32 bit; add sound & init system.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1148,7 +1148,7 @@ elseif(Haiku)
         src/detection/gtk_qt/gtk.c
         src/detection/host/host_windows.c
         src/detection/icons/icons_nosupport.c
-        src/detection/initsystem/initsystem_linux.c
+        src/detection/initsystem/initsystem_haiku.cpp
         src/detection/keyboard/keyboard_nosupport.c
         src/detection/libc/libc_nosupport.c
         src/detection/lm/lm_nosupport.c
--- a/src/detection/cpu/cpu_haiku.c
+++ b/src/detection/cpu/cpu_haiku.c
@@ -10,7 +10,7 @@ const char* ffDetectCPUImpl(FF_MAYBE_UNUSED const FFCPUOptions* options, FFCPURe
     if (get_system_info(&sysInfo) != B_OK)
         return "get_system_info() failed";
 
-    uint32_t topoNodeCount = 0;
+    uint32 topoNodeCount = 0;
     get_cpu_topology_info(NULL, &topoNodeCount);
     if (topoNodeCount == 0)
         return "get_cpu_topology_info(NULL) failed";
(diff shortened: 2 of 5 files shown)


CarterLi:
init of haiku isn't pid 1?

CarterLi:
Does Host module work for you?

mmuman:
> init of haiku isn't pid 1?

No, pid 1 is the kernel actually, and unlike BeOS, Haiku shares the thread and team(process) namespaces, so like now launch_daemon is 57.

> Does Host module work for you?

I seem to get some DMI info yes.

CarterLi:
> No, pid 1 is the kernel actually

In Haiku kernel is a userland process?

> I seem to get some DMI info yes.

Does it print SM2 or SM3?

CarterLi:
Does Haiku have any API / syscalls to query DMI data?

mmuman:
> In Haiku kernel is a userland process?

No but it is treated as a process regardless.

> Does it print SM2 or SM3?

SM2.

mmuman:
> Does Haiku have any API / syscalls to query DMI data?

No, I don't even think we use these in the kernel, unlike ACPI.

CarterLi:
Strangely it doesn't work for @Begasus. Maybe there are some bugs in SM3 related code

mmuman:
Or maybe it only works for 32bit?

CarterLi:
> > Does Haiku have any API / syscalls to query DMI data?
> 
> No, I don't even think we use these in the kernel, unlike ACPI.

I do find this one; doesn't seem to be useful though: https://github.com/haiku/haiku/blob/master/headers/private/drivers/smbios.h

CarterLi:
> Or maybe it only works for 32bit?

It works for me in my VM, 64bit. But it also uses SM2

CarterLi:
My compiler complains that variable mute is set but not used

mmuman:
> My compiler complains that variable mute is set but not used

I reused code from setvolume, I wasn't sure on how to report it, either ignore or report 0% as volume?

Begasus:
Probably related output in Terminal?
```
BParameter::GetValue parameter 'Mute' querying node 2, port 280 failed: General system error
```

mmuman:
Hmm it shouldn't print an error…

Begasus:
I get this on both arch's

mmuman:
It's odd, I wouldn't think the mixer would have different configurations.
Do we want to handle it or just remove that?

Begasus:
Best remove or disable it untill it could be figured out

mmuman:
I don't see why you get an error message, the code only enumerates parameters, and only reads the gain one for now…

Begasus:
> Hmm it shouldn't print an error…

Should problably mention it's not a build error, it shows up when running fastfetch in Terminal. :)

mmuman:
Yes but we don't try to get the value yet… :shrug:

Begasus:
Maybe showed up with the debug enabled?

mmuman:
Oh Id didn't know someone else wrote it already… :-D

mmuman:
Ah, now after rebasing I get the same error :sweat_smile:

mmuman:
Argh, sizeof(bool) is 1, not 4, which is why it complains. change isMute to int32.

Begasus:
That worked, error is gone now.

PR #1575, opened 2025-02-19
Battery (Haiku): handle empty slots

mmuman:
My ThinkPad has a second slot for external batteries.


CarterLi:
You should check `extended.last_full_charge == 0`. If true, just skip this battery.

mmuman:
I just copied what PowerStatus did but let's try that.

mmuman:
Actually `(uint32)-1` but yeah, should work.

PR #1580, opened 2025-02-22
Add support for Pisi Linux

keremgokcek:
Fixed Pisi Linux logo and added support for pisi package manager.


CarterLi:
Seems [Pisi Linux](https://github.com/pisilinux/PisiLinux) was dead long ago. The logo can be fixed since it's already there but the package manager isn't worth supporting.

keremgokcek:
It's not really dead, that's old repository. These are current repositories, packages are mostly up to date. Also last Pisi Linux release (Pisi Linux 2.4) was 4 months ago.

https://github.com/pisilinux/core
https://github.com/pisilinux/main
https://github.com/pisilinux/contrib

image

PR #1582, opened 2025-02-23
Logo (Builtin): Added Bedrock Linux small ASCII logo

TurtleMeds:
Added a bedrock linux ASCII logo for the small mode.
here is the ASCII art (without color codes) for reference:
```
    _________ 
   | __      |
   | \ \___  |
   |  \  _ \ |
   |   \___/ |
   |_________|

```


CarterLi:
I removed the internal padding. You should use `--logo-padding-left <NUM>` if needed

PR #1588, opened 2025-02-26
TerminalFont (Linux): support termite terminal

koraynilay:
Hi, I noticed it was missing font detection for the terminal `termite` (which I use) so I added it.

The config is simple ini-like config file.

Currently developed termite fork: https://github.com/aperezdc/termite.


CarterLi:
Another VTE based terminal. Okey...

PR #1595, opened 2025-03-03
src: logo: builtin: Add Common Torizon OS as a variant

microhobby:
There is another variant of Torizon OS called Common Torizon OS

### DIFF
--- a/src/logo/builtin.c
+++ b/src/logo/builtin.c
@@ -1093,6 +1093,16 @@ static const FFlogo C[] = {
         .colorKeys = FF_COLOR_FG_BLUE,
         .colorTitle = FF_COLOR_FG_WHITE,
     },
+    // Common Torizon
+    {
+        .names = {"common-torizon"},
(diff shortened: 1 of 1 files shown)


microhobby:
@CarterLi I do not have right now a build to test it on the hardware, I'm using information from a report:

image

And from the source code:

https://github.com/torizon/meta-toradex-torizon/blob/04e47c157cbe70d2147eb46e9892a221f1ef9a9a/conf/distro/common-torizon.conf#L5

CarterLi:
Please use `fastfetch -s os --format json`. The default output doesn't help.

CarterLi:
So it should be `common-torizon`
Note that putting `common-torizon` into `T` section wont work. It must be put into C section

microhobby:
@CarterLi so, let me see if I get it. It's better to use the distro id than the name on the `.names`?

microhobby:
@CarterLi added it to the C section as recommended, thanks for the review.

PR #1611, opened 2025-03-10
use `hyprctl version` instead of `Hyprctl --version` when detecting Hyprland version

### LINKED ISSUE #1610: [FEAT] use `hyprctl version`  as it is faster than `Hyprland --version`
### Description

in `src/detection/wm/wm_linux.c` run `hyprctl version`  instead of `Hyprland --version`

### Motivation

`hyprctl version` is 5-15x faster than `Hyprland --version` and prints the exact same output.
`hyprctl` requires Hyprland to be running to work, but if `wmName` is `Hyprland` then it should already be running.

### Additional context

```console
$ time Hyprland --version
Hyprland 0.45.2 built from branch v0.45.2-b at commit 12f9a0d0b93f691d4d9923716557154d74777b0a  ([gha] Nix: update inputs).
Date: 2024-11-19
Tag: v0.45.2, commits: 12f9a0d0b93f691d4d9923716557154d74777b0a
built against aquamarine 0.4.4


flags set:
debug


________________________________________________________
Executed in   88.75 millis    fish           external
   usr time   53.18 millis  676.00 micros   52.51 millis
   sys time   34.35 millis   55.00 micros   34.30 millis
```
```console
$ time hyprctl version
Hyprland 0.45.2 built from branch v0.45.2-b at commit 12f9a0d0b93f691d4d9923716557154d74777b0a  ([gha] Nix: update inputs).
Date: 2024-11-19
Tag: v0.45.2, commits: 12f9a0d0b93f691d4d9923716557154d74777b0a
built against aquamarine 0.4.4


flags set:
debug


________________________________________________________
Executed in    7.58 millis    fish           external
   usr time    2.73 millis    0.00 micros    2.73 millis
   sys time    4.52 millis  869.00 micros    3.65 millis
```

Image

Book-reader:
resolves #1610 

breaking changes:
 - fastfetch in a distrobox container will now report the host hyprland version instead of the container hyprland version.

before:
```console
$ time fastfetch -l none
 :  Hyprland 0.45.2 (Wayland)

________________________________________________________
Executed in   98.18 millis    fish           external
   usr time   47.13 millis  816.00 micros   46.31 millis
   sys time   48.31 millis  112.00 micros   48.20 millis
```
after:
```console
$ time ./build/fastfetch -l none
 :  Hyprland 0.45.2 (Wayland)

________________________________________________________
Executed in   23.55 millis    fish           external
   usr time    5.74 millis  786.00 micros    4.96 millis
   sys time   16.30 millis   58.00 micros   16.25 millis
```

### DIFF
--- a/src/detection/wm/wm_linux.c
+++ b/src/detection/wm/wm_linux.c
@@ -26,15 +26,15 @@ static bool extractHyprlandVersion(const char* line, FF_MAYBE_UNUSED uint32_t le
 static const char* getHyprland(FFstrbuf* result)
 {
     FF_STRBUF_AUTO_DESTROY path = ffStrbufCreate();
-    const char* error = ffFindExecutableInPath("Hyprland", &path);
-    if (error) return "Failed to find Hyprland executable path";
+    const char* error = ffFindExecutableInPath("hyprctl", &path);
+    if (error) return "Failed to find hyprctl executable path";
 
     if (ffBinaryExtractStrings(path.chars, extractHyprlandVersion, result, (uint32_t) strlen("v0.0.0")) == NULL)
         return NULL;
 
     if (ffProcessAppendStdOut(result, (char* const[]){
         path.chars,
-        "--version",
+        "version",
         NULL
     }) == NULL)
     { // Hyprland 0.46.2 built from branch v0.46.2-b at... long and multi line
@@ -43,7 +43,7 @@ static const char* getHyprland(FFstrbuf* result)
         return NULL;
     }
 
-    return "Failed to run command `Hyprland --version`";
+    return "Failed to run command `hyprctl version`";
 }
 
 static bool extractSwayVersion(const char* line, FF_MAYBE_UNUSED uint32_t len, void *userdata)


Book-reader:
It looks like this PR makes things significantly faster on NixOS, but there is no improvement / is slightly slower on Arch?

Book-reader:
Looks like there is a similar improvement on Arch, but the fastfetch in the Arch repos is somehow faster than the fastfetch built from source using the PKGBUILD in the arch repos.

image
the differing hyprland version is due to the distrobox container it is being run in, but the time should be accurate

src/detection/wm/wm_linux.c
```diff
@@ -26,15 +26,15 @@ static bool extractHyprlandVersion(const char* line, FF_MAYBE_UNUSED uint32_t le
 static const char* getHyprland(FFstrbuf* result)
 {
     FF_STRBUF_AUTO_DESTROY path = ffStrbufCreate();
-    const char* error = ffFindExecutableInPath("Hyprland", &path);
```
CarterLi, line 29: The version string should be extracted from Hyprland binary file directly by `ffFindExecutableInPath`, which is equivalent to `strings (which Hyprland) | rg 'v\d+\.\d+\.\d+'`. This is almost zero cost because it doesn't spawn processes at all.

If it doesn't work, fastfetch is either built without libelf support, or fastfetch fails to load `libelf.so`. You should check it

CarterLi:
I don't really like the change. As your comments read, `hyprctl` is not `Hyprland`. It's used as a fall back so I don't really care its performance.

src/detection/wm/wm_linux.c
```diff
@@ -26,15 +26,15 @@ static bool extractHyprlandVersion(const char* line, FF_MAYBE_UNUSED uint32_t le
...
-    if (error) return "Failed to find Hyprland executable path";
+    const char* error = ffFindExecutableInPath("hyprctl", &path);
+    if (error) return "Failed to find hyprctl executable path";
 
     if (ffBinaryExtractStrings(path.chars, extractHyprlandVersion, result, (uint32_t) strlen("v0.0.0")) == NULL)
```
CarterLi, line 32: The version string should be extracted from Hyprland binary file directly by `ffBinaryExtractStrings`, which is equivalent to `strings (which Hyprland) | rg 'v\d+\.\d+\.\d+'`. This is almost zero cost because it doesn't spawn processes at all.

If it doesn't work, fastfetch is either built without libelf support, or fastfetch fails to load `libelf.so`. You should check it

Book-reader:
looks like the problem is that on NixOS `Hyprland` is a shell script that wraps a differently named executable in a different location, so the first method fails.

Since it is a NixOS specific issue I'll close this & do something downstream instead.

PR #1622, opened 2025-03-18
Display (Linux): fix the error check for ffReadFileData

apocelipes:
Snipaste_2025-03-18_19-27-40

`ffReadFileData` returns -1 when error happens, so `!ffReadFileData(drmDir.chars, sizeof(buf), &buf) || buf != 'e'` becomes `false || buf != 'e'` and the `buf` was not initlialised.


CarterLi:
Good catch. Thanks.

PR #1683, opened 2025-04-10
Create ascii logo anduinos

dvorakchen:
AnduinOS is a custom Debian-based Linux distribution that aims to facilitate developers transitioning from Windows to Ubuntu by maintaining familiar operational habits and workflows.


Anduin2017:
Maybe you also need to update this file?

https://github.com/fastfetch-cli/fastfetch/blob/master/src/logo/builtin.c

PR #1695, opened 2025-04-17
T2/SDE: upstreaming downstream's logo change

x-zvf:
T2/SDE port of fastfetch applies two patches, improving the distribution's logo and adding alternative names. I see no point not having these upstream.

Downstream's patches are found here (SVN mirror):
https://github.com/rxrbln/t2sde/tree/main/package/contrib/fastfetch


CarterLi:
Instead of removing the old one, it should be good to rename it to `T2_small`

x-zvf:
> Instead of removing the old one, it should be good to rename it to `T2_small`

Good idea, done.

PR #1735, opened 2025-05-05
DisplayServer (Linux): fix xcb_connect error handling

apocelipes:
image
image

[doc](https://xcb.freedesktop.org/manual/group__XCB__Core__API.html)

We should use  `xcb_connection_has_error` to check whether errors have happened.


CarterLi:
Make sense. Thanks.

PR #1756, opened 2025-05-15
Logo (Builtin): Add Kylin Logo

zx900930:
image


CarterLi:
We don't accept ASCII logos with ANSI escape codes embeded. Sorry.

zx900930:
> We don't accept ASCII logos with ANSI escape codes embeded. Sorry.
OK, I'll fix that.

PR #1802, opened 2025-06-11
Fix NetBSD CPU detection for PPC

techflashYT:
Some devices store their CPU name in hw.model, so this patch adds support for checking that as well before giving up.  Tested on NetBSD 10.1 evbppc/wii, and allows it to detect the CPU as "750 (Revision 2.0)".


techflashYT:
Oh yikes, my branch accidentally included some old Linux stuff I had lying around from last time, fixing that, one moment.

Edit: Should be fixed now (oops!)

PR #1803, opened 2025-06-12
TerminalShell (Linux): better support for Ghostty

apocelipes:
Changes:

- Remove the suffix from the Ghostty version string on Arch Linux, which is not present in other distributions or macOS.

Ghostty does not support Windows (still WIP at an early stage), so only *nix parts modified.

### DIFF
--- a/src/detection/terminalshell/terminalshell.c
+++ b/src/detection/terminalshell/terminalshell.c
@@ -576,6 +576,25 @@ static bool getTerminalVersionZed(FFstrbuf* exe, FFstrbuf* version)
     return true;
 }
 
+static bool getTerminalVersionGhostty(FFstrbuf* exe, FFstrbuf* version)
+{
+    const char* env = getenv("TERM_PROGRAM_VERSION");
+    if (env)
+    {
+        ffStrbufAppendS(version, env);
+        // 1.1.3-arch1
+        ffStrbufSubstrBeforeFirstC(version, '-');
+        if(version->length > 0) return true;
+    }
+
+    if(!getExeVersionRaw(exe, version)) return false;
+    // Ghostty 1.1.3-arch1\n\nVersion\n...
+    ffStrbufSubstrBeforeFirstC(version, '\n'); // Ghostty 1.1.3-arch1
+    ffStrbufSubstrAfterFirstC(version, ' ');   // 1.1.3-arch1, only Arch Linux has the suffix '-arch1'
+    ffStrbufSubstrBeforeFirstC(version, '-');  // 1.1.3
+    return version->length > 0;
+}
+
 #ifndef _WIN32
 static bool getTerminalVersionKitty(FFstrbuf* exe, FFstrbuf* version)
 {
@@ -872,6 +891,9 @@ bool fftsGetTerminalVersion(FFstrbuf* processName, FF_MAYBE_UNUSED FFstrbuf* exe
 
     if(ffStrbufStartsWithIgnCaseS(processName, "zed"))
         return getTerminalVersionZed(exe, version);
+
+    if(ffStrbufStartsWithIgnCaseS(processName, "ghostty"))
+        return getTerminalVersionGhostty(exe, version);
 
     #if __HAIKU__
     if(ffStrbufEqualS(processName, "Terminal"))


src/detection/terminalshell/terminalshell_linux.c
```diff
@@ -226,6 +226,13 @@ static void getTerminalFromEnv(FFTerminalResult* result)
         getenv("ALACRITTY_WINDOW_ID") != NULL
     ) term = "Alacritty";
 
+    //Ghostty
+    else if(
+        getenv("GHOSTTY_RESOURCES_DIR") != NULL ||
+        getenv("GHOSTTY_SHELL_INTEGRATION_NO_SUDO") != NULL ||
+        getenv("GHOSTTY_BIN_DIR") != NULL
+    ) term = "Ghostty";
```
CarterLi: Ghostty reports `$TERM_PROGRAM`, so additional env-var checks are not needed.

apocelipes: Yep. Removed.

apocelipes: Yep. Removed.

src/detection/terminalshell/terminalshell.c
```diff
@@ -576,6 +576,25 @@ static bool getTerminalVersionZed(FFstrbuf* exe, FFstrbuf* version)
...
+    const char* env = getenv("TERM_PROGRAM_VERSION");
+    if (env)
+    {
+        ffStrbufAppendS(version, env);
+        // 1.1.3-arch1
+        ffStrbufSubstrBeforeFirstC(version, '-');
+        if(version->length > 0) return true;
+    }
+
+    if(!getExeVersionRaw(exe, version)) return false;
```
CarterLi, line 590: I actually dont think this is needed. As $TERMINAL_PROGRAM_VERSION exists.
We rely on it for some other terminals too, such as Wezterm, iTerm and warp.

Note we always add another detection method after ffBinaryExtractStrings usage, as ffBinaryExtractStrings relies on the internal implementation of terminals which is not very reliable. While $TERMINAL_PROGRAM_VERSION is considerred a public API; if the variable is not there it is a terminal bug.

src/detection/terminalshell/terminalshell.c
```diff
@@ -576,6 +576,25 @@ static bool getTerminalVersionZed(FFstrbuf* exe, FFstrbuf* version)
     return true;
 }
 
+static bool getTerminalVersionGhostty(FFstrbuf* exe, FFstrbuf* version)
+{
+    const char* env = getenv("TERM_PROGRAM_VERSION");
+    if (env)
+    {
+        ffStrbufAppendS(version, env);
+        // 1.1.3-arch1
```
CarterLi, line 585: FYI fedora version also includes `.fc42`. Lets keep as it was.

apocelipes:
Ubuntu:
ubuntu
ubuntu2

Arch Linux:
archlinux1
archlinux2

apocelipes:
Rebased. Commit message has been updated.

CarterLi:
Hello. Please reply my comments first.

apocelipes:
> Hello. Please reply my comments first.

I'm sorry, I didn't see it, could you quote-reply the comment and @ me? 

I only saw this and replied:
image

apocelipes:
> > Hello. Please reply my comments first.
> 
> I'm sorry, I didn't see it, could you quote-reply the comment and @ me?
> 
> I only saw this and replied: image

@CarterLi

src/detection/terminalshell/terminalshell.c
CarterLi: @apocelipes

src/detection/terminalshell/terminalshell.c
CarterLi: image

For nightly version, `ghostty --version` includes the branch name and commit hash.

CarterLi: The kernel version contains archlinux specific suffix `-arch1-1`. I don't think it's problematic. Do you?

apocelipes:
@CarterLi So, where is your comment? Could you paste a screenshot like what I did?

CarterLi (changes requested):
Please review these comments

CarterLi:
Can you see it now?

apocelipes:
> Can you see it now?

Thanks, I can see it now. Maybe it's Github's bug that it didn't display to me.

apocelipes:
The version suffix doesn't indicate a change in the program code, sometimes the suffix changes simply by packager recompiling it. Many other programs don't have the suffix in their versions output, such as bash (usually this is a side effect of the formatting), so I'm considering whether or not to remove ghostty's version suffix just the same.

But since Fedora also has a version suffix, I think it's okay to keep it.

I'll close this PR. Thanks for your review.

PR #1835, opened 2025-07-05
Added support for WM & DE Detection on Android/Termux

Usinganame:
Added WM & DE Detection to Android as it work the same was as on other *nix systems. I don't know about wayland, as sway doesn't like my existing i3 config so I can't get a terminal open, but it should work. Tested with i3 and Xfce

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,6 +12,7 @@ set(PROJECT_LICENSE "MIT license")
 ###################
 # Target Platform #
 ###################
+
 if(ANDROID)
     set(LINUX FALSE)
 elseif("${CMAKE_SYSTEM_NAME}" STREQUAL "Linux")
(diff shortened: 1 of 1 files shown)


CarterLi:
I don't think Android/Termux has WM, DE and stuff. For proot distros, you should install the Linux version ( not officially supported ).

Usinganame:
It actually supports X11 very well, even being able to get a full DE without Proot. 
https://wiki.termux.com/wiki/Graphical_Environment
Screenshot_20250705_044511_TermuxX11

robertkirkman:
Hello, we can see that Usinganame has posted a screenshot, but also that you decide not to enable this for Android, that is ok, for Termux there is a PR so there will be a discussion about whether it's feasible to enable the WM and DE detection in `fastfetch` on Termux. If it's decided that the PR will be feasible, a package might be added to Termux that has these features enabled, and then at least some users will be able to test it and give feedback about how stable `fastfetch` is for detecting DE and WM on Termux.

PR #1858, opened 2025-07-21
Fix qi packages detection.

lidgnulinux:
This PR fix qi packages detection, /var/qi/installed_packages.list is no more needed and we read /usr/pkg directory directly.


CarterLi:
Hello. pkgsrc ( a package manager used by NetBSD, but also support Linux ) also use `/usr/pkg`. In addition, `/usr/pkg` is commonly used by BSD systems.

https://github.com/fastfetch-cli/fastfetch/blob/7566f10bb6463bee127434f5d18892fa2017a507/src/detection/packages/packages_linux.c#L581

In this case, I'm afraid that I can't support qi any more.

> we read /usr/pkg directory directly

Are you one of the maintainers of qi?

lidgnulinux:
Qi seems to be little different. Qi install packages by extracting packages into /usr/pkg/name_of_package. 

> Are you one of the maintainers of qi?

No, I'm not, the dragora folks (people who use qi) seems not care. If the qi package detection support will be dropped, I'll maintain the patch for my self.

lidgnulinux:
I was the one who previously suggest to use /var/qi/installed_packages.list months ago. Officially, /var/qi/installed_packages.list doesn't exist. I just hack the qi package manager to make that file so I can read what packages are installed.

CarterLi:
/usr/pkg is a very common path name, and has been used by other package managers.

I dont think counting elements in /usr/pkg is a good idea. If existing method doesnt work any more, I will delete it.

Thanks for reporting!

PR #1865, opened 2025-07-25
Fix: named tags in format strings not resolving correctly

SorYoshino:
After completing the tests, all modules—except for `btrfs`, `gamepad`, `media`, `player`, `poweradapter`, and `zpool`—were found to parse correctly. Issues were identified only in the `version` and `users` modules, both of which have been fixed and now produce the expected output.


CarterLi:
Thanks.

PR #1880, opened 2025-08-02
add ability to count aur pkgs

pheiduck:


CarterLi:
After applying your patch, the execution time of packages module increases from 1.043ms to 305.441ms, which is unacceptable. In addition, it prints `1657 (pacman)13 (aur),`, I think it's badly formatted.

Besides, I don't think foreign packages are always AUR packages. They can be packages built locally with `PKGBUILD` or installed directly with `.pkg.tar.zst` files.

PR #1882, opened 2025-08-03
Fix lliurex detection

juanma1980:
Hi. Some time ago I did a PR for adding support for the Lliurex distribution. After the PR was accepted with changes I didn't test them (sorry for that) because i was facing some troubles and forgot about this but as lliurex is a mix of Ubuntu LTS and Neon the detection based on XDG_CONFIG_DIRS was failing because conflicted with kubutnu as both use plasma as desktop.
This patch fixes this problem checking for the presence of a lliurex-specific file (/etc/lliurex-cdd-version) and getting the version from another specific command of llurex.
I've included the detection code just before the tests for XDG_CONFIG_DIRS because xdg_config_dirs in lliurex also contains "plasma" so any detection mechanism for lliurex must run before the check for kubuntu; also the old detection code has been removed.
Thanks!

### DIFF
--- a/src/detection/os/os_linux.c
+++ b/src/detection/os/os_linux.c
@@ -88,6 +88,20 @@ FF_MAYBE_UNUSED static void getUbuntuFlavour(FFOSResult* result)
         ffStrbufSetStatic(&result->id, "vanilla");
         ffStrbufSetStatic(&result->idLike, "ubuntu");
     }
+    if (ffPathExists("/etc/lliurex-cdd-version", FF_PATHTYPE_FILE))
+	{
+        ffStrbufSetStatic(&result->name, "LliureX");
+        ffStrbufSetStatic(&result->id, "lliurex");
+        ffStrbufClear(&result->version);
+        if (ffProcessAppendStdOut(&result->version, (char* const[]) {
+            "/usr/bin/lliurex-version",
+            NULL,
+        }) == NULL) // 8.2.2
+            ffStrbufTrimRightSpace(&result->versionID);
+        ffStrbufSetF(&result->prettyName, "LliureX %s", result->version.chars);
+        ffStrbufSetStatic(&result->idLike, "ubuntu");
+        return;
+	}
 
     if(ffStrContains(xdgConfigDirs, "kde") || ffStrContains(xdgConfigDirs, "plasma") || ffStrContains(xdgConfigDirs, "kubuntu"))
     {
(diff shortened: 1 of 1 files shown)


src/detection/os/os_linux.c
```diff
@@ -89,6 +89,21 @@ FF_MAYBE_UNUSED static void getUbuntuFlavour(FFOSResult* result)
...
+    if (ffPathExists("/etc/lliurex-cdd-version", FF_PATHTYPE_FILE))
+	{
+        ffStrbufSetStatic(&result->name, "LliureX");
+        ffStrbufSetStatic(&result->id, "lliurex");
+        ffStrbufClear(&result->version);
+        if (ffProcessAppendStdOut(&result->version, (char* const[]) {
+            "/usr/bin/lliurex-version",
+            NULL,
+        }) == NULL) // 8.2.2
+            ffStrbufTrimRightSpace(&result->versionID);
```
CarterLi: You appended the version string to `&result->version`, but triming `&result->versionID` here?

CarterLi: And you put executable binaries to `/etc/`? Shouldn't it be a regular text file?

src/detection/os/os_linux.c
```diff
@@ -89,6 +89,21 @@ FF_MAYBE_UNUSED static void getUbuntuFlavour(FFOSResult* result)
...
+        ffStrbufSetF(&result->prettyName, "LliureX %s", result->version.chars);
```
CarterLi, line 102: What if ffProcessAppendStdOut failed?

src/detection/os/os_linux.c
juanma1980: Hi, the trim was left by mistake.
For the another question there's no binary at /etc. The check for /etc/lliurex-cdd-version is for a regular txt file, the binary to get the lliurex version is at /usr/bin/lliurex-version. First checks for the file, then (if file exists) gets the version through a binary. I could check the binary itself directly but it will fail if the aforementioned file didn't exist so I believe is better to check for the file.

p.s. I've made another commit fixing the trim

src/detection/os/os_linux.c
juanma1980: lliurex-version is a basic system command in lliurex, if it fails then something is really wrong, it wold be as if command lsb_release fails. If you want I can add any behaviour on fail but the tests I did seemed to be ok without any issue at all.

Cheers

PR #1888, opened 2025-08-10
Show brand of Xserver in WM section

aequabit:
Putting it in square brackets next to the WM seemed like the most logical choice after looking at how icon and font information is displayed. If you want any changes made, just tell me and I'll take care of it.

image

### DIFF
--- a/src/detection/displayserver/displayserver.c
+++ b/src/detection/displayserver/displayserver.c
@@ -61,6 +61,7 @@ const FFDisplayServerResult* ffConnectDisplayServer()
         ffStrbufInit(&result.wmProtocolName);
         ffStrbufInit(&result.deProcessName);
         ffStrbufInit(&result.dePrettyName);
+        ffStrbufInit(&result.serverVendor);
         ffListInit(&result.displays, sizeof(FFDisplayResult));
         ffConnectDisplayServerImpl(&result);
     }
--- a/src/detection/displayserver/linux/xcb.c
+++ b/src/detection/displayserver/linux/xcb.c
@@ -6,6 +6,7 @@
 #include "common/time.h"
 #include "util/edidHelper.h"
 #include "util/mallocHelper.h"
+#include "util/stringUtils.h"
 
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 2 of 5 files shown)


CarterLi:
What does it print on original X11?

src/detection/displayserver/displayserver.c
```diff
@@ -61,6 +61,7 @@ const FFDisplayServerResult* ffConnectDisplayServer()
         ffStrbufInit(&result.wmProtocolName);
         ffStrbufInit(&result.deProcessName);
         ffStrbufInit(&result.dePrettyName);
+        ffStrbufInit(&result.serverVendor);
```
CarterLi, line 64: So this is only used for X11. I think it should be ok to append the string to `wmProtocolName`. Like `X11-XLibre`

src/detection/displayserver/linux/xcb.c
```diff
@@ -86,6 +93,25 @@ static void xcbDetectWMfromEWMH(XcbPropertyData* data, xcb_connection_t* connect
...
+
+    //Why are xcb property strings not null terminated???
+    void* replyValue = malloc((size_t)length + 1);
+    memcpy(replyValue, data->ffxcb_setup_vendor(setup), (size_t) length);
+    ((char*) replyValue)[length] = '\0';
+
+    if (ffStrSet(replyValue))
+        ffStrbufSetS(&result->serverVendor, replyValue);
+
+    free(replyValue);
```
CarterLi: `ffStrbufSetNS(&result->serverVendor, (uint32_t) length, data->ffxcb_setup_vendor(setup))`

src/detection/displayserver/displayserver.c
aequabit: I think differentiating between the protocol and the implementation detail is important (as it's done for fonts already), even if it means allocating a few more bytes.

src/detection/displayserver/linux/xcb.c
aequabit: Done.

src/detection/displayserver/displayserver.c
CarterLi: 1. Most, if not all, distros still use the original X11 impl by default. So syou get `The X.Org Foundation` in most cases.
2. More and more distros are moving to Wayland. For Wayland it's not used at all.

I still think a separate field is not necessary. I'd like to implement this:

1. If `xcb_setup_vendor` reports `The X.Org Foundation`, set protocol name to `X11`
2. Otherwise, set protocol name to whatever `xcb_setup_vendor` reports ( in your case, `XLibre` )

aequabit: I really should've tested this on Xorg before submitting, I assumed the vendor string to be "Xorg". Sorry about that.

What do you think about `WM: KWin (X11) [Xorg]` and `WM: KWin (X11) [XLibre]`? Or maybe `WM: KWin (X11/Xorg)` and  `WM: KWin (X11/XLibre)` ?

CarterLi: I prefer `WM: KWin (X11)` and `WM: KWin (XLibre)`, so if a user uses the original X11 nothing changes.

CarterLi:
> I prefer `WM: KWin (X11)` and `WM: KWin (XLibre)`

I will do this change myself.

PR #1923, opened 2025-08-20
Fix Anduinos support

Funasitien:
This PR try to fix AnduinOS not detected, and add a better lokking anduinos ascii art

image


CarterLi:
An acceptable logo must

1. Be smaller than 50x20 characters (width x height)
2. Contain no internal paddings

And I don't think your fix actually fixes anything, as OS names are always matched case-insensitively

PR #1931, opened 2025-08-24
Add GNU/Hurd support

### LINKED ISSUE #1895: [FEAT] Add support for Debian GNU/Hurd
### Description

[Debian GNU/Hurd 2025 Released With Completed 64-bit Support, Rust Ported](https://www.phoronix.com/news/Debian-GNU-Hurd-2025).

Currently, this is what I get running `cmake`:

```
Unsupported platform: GNU
```

Image

### Motivation

It'd be nice to be able to run `fastfetch` on Debian GNU/Hurd.

### Additional context

Here is a screenshot of `uname -a` and `/etc/os-release`:

Image

CarterLi:
Based on work of https://codeberg.org/Yelninei/fastfetch

image

Fixes #1895

@yelninei @osalbahr Please test!


osalbahr:
It works. Thx!

image

src/common/netif/netif_gnu.c
```diff
@@ -0,0 +1,35 @@
...
+
+#include <net/if.h>
+#include <stdio.h>
+
+#define FF_STR_INDIR(x) #x
+#define FF_STR(x) FF_STR_INDIR(x)
+
+bool ffNetifGetDefaultRouteImplV4(FFNetifDefaultRouteResult* result)
+{
+    // Based on netif_linux.c before 5e770dc8b019702ca458cc0cad46161ebec608a4
```
yelninei: Is my comment here still necessary? It feels a bit irrelevant when it is in fastfetch proper and not in my fork

src/detection/initsystem/initsystem_linux.c
```diff
@@ -84,6 +84,20 @@ const char* ffDetectInitSystem(FFInitSystemResult* result)
                 ffStrbufSubstrAfterLastC(&result->version, ' ');
             }
         }
+        else if (ffStrbufEqualS(&result->name, "guile"))
```
yelninei, line 87: I would not add this as it is just misleading and wrong

Instead it would be better to try to support shepherd instead but this can be done in a followup at a later time

src/common/processing_linux.c
```diff
@@ -465,11 +465,30 @@ const char* ffProcessGetBasicInfoLinux(pid_t pid, FFstrbuf* name, pid_t* ppid, i
...
     else
     {
+        #ifndef __GNU__
         snprintf(procFilePath, sizeof(procFilePath), "/proc/%d/comm", (int)pid);
         ssize_t nRead = ffReadFileBuffer(procFilePath, name);
         if(nRead <= 0)
             return "ffReadFileBuffer(/proc/pid/comm, name) failed";
         ffStrbufTrimRightSpace(name);
+        #else
+        // No /proc/1/comm on Hurd so read /proc/1/stat again
```
yelninei: I am not really happy with this and feels a bit overcomplicated just to get the name of pid1, especially when the initsystem module will set it to basename of /proc/1/cmdline later.

src/detection/packages/packages.h
```diff
@@ -55,7 +55,7 @@ const char* ffDetectPackages(FFPackagesResult* result, FFPackagesOptions* option
 bool ffPackagesReadCache(FFstrbuf* cacheDir, FFstrbuf* cacheContent, const char* filePath, const char* packageId, uint32_t* result);
 bool ffPackagesWriteCache(FFstrbuf* cacheDir, FFstrbuf* cacheContent, uint32_t num_elements);
 
-#if defined(__linux__) || defined(__APPLE__)
+#if defined(__linux__) || defined(__APPLE__) || defined(__GNU__)
```
yelninei, line 58: I guess that works as well.

I didnt want to do this because Nix does not support hurd

src/detection/initsystem/initsystem_linux.c
CarterLi: I don't use guix. Google said that guile is a programing language. Is guile here the name of guile interpreter process?

src/common/processing_linux.c
CarterLi: It's common code. It should be implemented correctly regardless it's used by initsystem or other module.

CarterLi: Just realized that a process name can contain white spaces and brackets. I tested this case and I found that Linux does not encode special characters in `stat` at all.

```
$ cat stat
8759 (
te st)) S 2766 8759 2766 34816 8759 4194304 57 0 0 0 0 0 0 0 20 0 1 0 221950 851968 98 18446744073709551615 187650765619200 187650765621512 281474827435536 0 0 0 0 0 0 1 0 0 17 2 0 0 0 0 0 187650765749672 187650765750280 187651010842624 281474827439786 281474827439796 281474827439796 281474827440110 0
```

Need to fix this case.

src/detection/initsystem/initsystem_linux.c
yelninei: Yes, the main shepherd program is a guile script that starts with

```
#!/path/to/guile --no-autocompile

....
```

Right now the guile version is being detected which is not the version of the initsystem.
The correct path to the shepherd executable is in `/proc/1/cmdline` but it is the string after the first 2 null bytes.

There is also a minor problem that on non-linux systems `shepherd --version` has an additional extra line in the output:

```
Linux:
shepherd --version
shepherd (GNU Shepherd) 1.0.6

Hurd:
shepherd --version
System lacks support for 'signalfd'; using fallback mechanism.
shepherd (GNU Shepherd) 1.0.6
```

I guess this can be detected by checking if the line starts with `shepherd` and skipping the first `\n`if it does not.

I'll see if I can make this work

CarterLi: We did similar special handling for python. Perhaps we can use the same logic.

CarterLi:
Hello @hiagofranco

As we introduced and tested GNU Hurd support on Debian, can we add fastfetch to Debian Hurd repo?

hiagofranco:
Hi @CarterLi,

Sure, I will check how I can proceed with that and get back to you guys with any updates.

Thanks

osalbahr:
`fastfetch` is now in Debian GNU/Hurd! Thx everyone.

Screenshot 2025-10-18 at 1 55 10 PM

hiagofranco:
@osalbahr, yes =) Actually when I did the upload I forgot to post a message here, sorry about that.

Enjoy

Issue #1974, opened 2025-09-22
[BUG] Generic Ubuntu logo when `XDG_CONFIG_DIRS` is not set for Ubuntu flavor (at least Rhino Linux)


osalbahr:
### Read the FAQ first

- [x] I have checked the FAQ but the issue is not covered

### Description

Ubuntu flavor detection is broken in the Rhino Linux container (`ghcr.io/rhino-linux/docker`) when `XDG_CONFIG_DIRS` is not set (it's not set by default). Is this expected behavior?

I did some investigation (spamming `puts`), and the culprit is here:

https://github.com/fastfetch-cli/fastfetch/blob/de9229697a8ac1712b45bb94543e18b451711312/src/detection/os/os_linux.c#L64-L66

(I don't actually use Rhino Linux, just something I noticed and thought I'd point it out)

### Version used

fastfetch 2.52.0-30 (aarch64)

### Bug prevalence

Always

### Regression

Not sure

### Installation

Built from source

### Screenshots

<img width="1018" height="721" alt="Image" src="https://github.com/user-attachments/assets/08d63886-deff-4219-b740-de274ff9dd61" />

### Configuration

```jsonc

```

### Image protocol

I don't know

### Terminal

Terminal (macOS)

### Image tried

N/A

### Features built-in

```text
threads
libzfs
linux/videodev2
linux/wireless
```

osalbahr:
I'll draft a PR moving the if statement down a bit (I only thought of this after creating the issue).

PR #1975, opened 2025-09-22
OS (Linux): Better Ubuntu flavor detection

osalbahr:
- Fixes https://github.com/fastfetch-cli/fastfetch/issues/1974


CarterLi:
Make sense to me. Thanks.

Issue #868, opened 2024-05-04
[SECURITY] out-of-bound in commandoption.c


RootUp:
### Summary

While fuzzing `fastfetch` (4175dfdc3a9990060c9d7681da86702d2639b3ad) it was found that the application suffers from out-of-bound due to lack of input validation, allowing application to crash via a crafted configuration files leading to denial or service or code execution.

### ASAN

```
==3290000==ERROR: UndefinedBehaviorSanitizer: SEGV on unknown address 0x007900000078 (pc 0x00000042cd55 bp 0x7fffffffe330 sp 0x7fffffffe090 T3290000)
==3290000==The signal is caused by a READ memory access.
    #0 0x42cd55 in ffParseModuleOptions /fastfetch/src/common/commandoption.c:16:77
    #1 0x428d39 in parseOption /fastfetch/src/fastfetch.c:737:9
    #2 0x42b86c in parseConfigFile /fastfetch/src/fastfetch.c:397:13
    #3 0x427bc6 in optionParseConfigFile /fastfetch/src/fastfetch.c:514:47
    #4 0x427bc6 in parseCommand /fastfetch/src/fastfetch.c:662:9
    #5 0x4278d3 in parseArguments /fastfetch/src/fastfetch.c:795:13
    #6 0x427014 in main /fastfetch/src/fastfetch.c:874:5
    #7 0x7ffff7c4e082 in __libc_start_main /build/glibc-e2p3jK/glibc-2.31/csu/../csu/libc-start.c:308:16
    #8 0x406bed in _start (/fastfetch/build/fastfetch+0x406bed)

UndefinedBehaviorSanitizer can not provide additional info.
SUMMARY: UndefinedBehaviorSanitizer: SEGV /fastfetch/src/common/commandoption.c:16:77 in ffParseModuleOptions
==3290000==ABORTING
```

### Code Snippet

https://github.com/fastfetch-cli/fastfetch/blob/dev/src/common/commandoption.c#L16

```C
bool ffParseModuleOptions(const char* key, const char* value)
{
    if (!ffStrStartsWith(key, "--") || !isalpha(key[2])) return false;

    for (FFModuleBaseInfo** modules = ffModuleInfos[toupper(key[2]) - 'A']; *modules; ++modules)
    {
        FFModuleBaseInfo* baseInfo = *modules;
        if (baseInfo->parseCommandOptions(baseInfo, key, value)) return true;
    }
    return false;
}

```

[...]

PR #1989, opened 2025-10-02
Replace Manual Alphanumeric Check

TAOFOR4:
Instead check if character is alphabet letter or number manually, we can utilize functions from C standard library.


CarterLi:
ctype functions are locale-dependent, which can cause problems. See #868

PR #2009, opened 2025-10-15
Don't flush terminal input when querying checking window size

krobelus:
As reported in https://github.com/MidnightCommander/mc/issues/4789,
after Midnight Commander starts a shell, it writes some characters
to the shells stdin, to initialize it.

Since mc doesn't know when shell initialization
(e.g. .bashrc), it doesn't wait for that.
Hence mc's commands race with bashrc.

This surfaces reliably when running fastfetch from .bashrc with a *.png
logo.  Writing the logo involves a call to ffDetectTerminalSize(),
which queries /dev/tty.

It first flushes (TCSAFLUSH) the input since a0705197 (IO: disallow
input echo, 2024-07-31), which removes the input queued by mc.

Then it reads from the TTY 1024 bytes at a time.
This is also bad because it might swallow key presses
made by the user.

So maybe
1. Don't flush.
2. Reduce the likelihood of swallowing users's key presses by reading
   only one byte at a time.

Note that with or without TCSAFLUSH,
our query responses might be interspersed with keyboard input.
Though I guess TCSAFLUSH makes this less likely.

### DIFF
--- a/src/common/io/io_unix.c
+++ b/src/common/io/io_unix.c
@@ -177,7 +177,11 @@ static int ftty = -1;
 static struct termios oldTerm;
 void restoreTerm(void)
 {
-    tcsetattr(ftty, TCSAFLUSH, &oldTerm);
+    tcsetattr(ftty, TCSANOW, &oldTerm);
+}
+
+static size_t min(size_t left, size_t right) {
+    return left < right ? left : right;
 }
 
 const char* ffGetTerminalResponse(const char* request, int nParams, const char* format, ...)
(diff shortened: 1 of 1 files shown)


CarterLi:
Does it really solve your issue? Even if I don't flush the input, I still have to read and ignore the input before I successfully find the terminal response. The input queued by other programs will be effectively removed just like what `TCSAFLUSH` does.

CarterLi:
`/dev/tty` is a shared resource. You can't assume that it's not touched by other programs. I don't know why do you want to write stdin. If you want to communicate with a subprocess, you should better use other IPC methods.

Issue #2027, opened 2025-10-26
[FEAT] Support SDKMAN! Package Count


iskandergaba:
### Description

Add [SDKMAN!](https://sdkman.io) package count to the output.

### Motivation

I think this falls under the same category as other package managers. SDKMAN! is popular and widely used across multiple Unix platforms to install system-wide SDKs and frameworks.

### Additional context

This should be relatively easy and useful for those using SDKMAN! to install SDKs and packages. Here's how to detect the count:
1. Get the candidates' directory from the variable `$SDKMAN_CANDIDATES_DIR`. If it is empty, then SDKMAN! is not installed.
2. Count the number of **_non-empty_** directories inside `$SDKMAN_CANDIDATES_DIR`. SDKMAN! can leave empty directories for uninstalled SDK candidates, so it is important to count the non-empty directories only (i.e., a bit more involved than a simple `getNumElements` call).

Note that this counts every SDK or framework installed as 1 (even if there are multiple versions installed for it), which is, in my opinion, the correct thing to do.

CarterLi:
I don't use the package manager. PR is welcome.

iskandergaba:
Alright, I will give it a try.

Issue #1048, opened 2024-06-26
[FEAT] Supports for npm global packages


nolddor:
## Wanted features:
https://nodejs.org/en/learn/getting-started/an-introduction-to-the-npm-package-manager
> **npm** is the standard package manager for Node.js.

Would be great if `packages` module from **fastfetch** add support for global packages installed system wide using de facto node.js package manager.

## Motivation:
Same as `apt` or `snap` I personally use `npm` eventually to add system-wide packages. By running the `npm install <package> -g` command. As for now, I have no way to check how many npm global packages there are on my system.

Current workaround is to use the `command` **fastfetch** module as per sample below:
```
    {
      "type": "command",
      "text": "ls -l \"$(npm root -g)\" | grep -c ^d |  awk '{print $1 \" (npm)\"}'",
      "key": "Packages"
   }
```

CarterLi:
Currently we have no plan for programming language specific package managers. I think the command module is enough for your personal use.

> de facto node.js package manager

There are also `yarn global` and `pnpm -g`

nolddor:
Sounds fair. Thx for your time checking this request!

PR #2033, opened 2025-10-29
Support SDKMAN! Package Count

iskandergaba:
This Pull Request fulfills #2027.

_Disclaimer_: I have used generative AI to help me implement this feature. I have, however, had to thoroughly review, optimize, and fix the code myself.

Below is a screenshot of the result. Note that some data has been redacted from the screenshot.

screenshot

Let me know if anything is missing.


CarterLi:
Does SDKMAN supports java SDKs only? [We don't plan to add support for language specific package managers.](https://github.com/fastfetch-cli/fastfetch/issues/1048#issuecomment-2192867358)

iskandergaba:
From their website, I quote:
> Install Software Development Kits for the JVM such as Java, Scala, Kotlin, and Groovy. Ant, Gradle, Grails, Maven, SBT, Spark, Spring Boot, Vert.x and many others also supported.

So, technically no. It is not only Java Development Kits (JDKs), but rather SDKs for Java Virtual Machine (JVM) and then some more. For example:
1. Apache Spark, for example, is built using Scala (a JVM functional programming language). While the engine itself is written in Scala, it offers high-level APIs in multiple programming languages, including Java, Python (via PySpark), R, and Scala, allowing developers to write applications in their preferred language.
2. Apache ActiveMQ is primarily written in Java and is a Java-based message broker developed by the Apache Software Foundation. While the core broker is Java-based, Active's supports clients written in a wide range of programming languages, including JavaScript, C, C++, Python, .NET, Ruby, PHP, and others.

> We don't plan to add support for language specific package managers.

The Java equivalent for `npm` is not `SDKMAN!`. It would be rather `maven` and `gradle`, which happen to be installable via `apt` as well as `SDKMAN!`.

My point is: This is not to install Java project libraries like `npm` or `pip`.

I laid my case. It is up to the maintainers and the community from here.

PR #2051, opened 2025-11-13
Zpool: add NetBSD support

uzuki314:
This PR adds NetBSD support to the zpool module.

netbsd-fastfetch-zpool

### DIFF
--- a/src/detection/zpool/zpool_linux.c
+++ b/src/detection/zpool/zpool_linux.c
@@ -109,9 +109,53 @@ const char* ffDetectZpool(FFlist* result /* list of FFZpoolResult */)
 
 #else
 
+#include "common/processing.h"
+#include "util/path.h"
+
 const char* ffDetectZpool(FF_MAYBE_UNUSED FFlist* result)
 {
-    return "Fastfetch was compiled without libzfs support";
+    FF_STRBUF_AUTO_DESTROY path = ffStrbufCreate();
+    const char* error = ffFindExecutableInPath("zpool", &path);
+    if (error) return "Failed to find zpool executable path";
+
+    FF_STRBUF_AUTO_DESTROY output = ffStrbufCreate();
+    if (ffProcessAppendStdOut(&output, (char* const[]){
+        path.chars,
+        "get",
+        "-Hp",
+        "health,version,size,allocated,fragmentation",
+        NULL
+    }) == NULL)
+    {
+        if (output.length == 0)
+            return NULL;
+
+        char* line = NULL;
+        size_t len = 0;
+        while (ffStrbufGetline(&line, &len, &output))
+        {
+            FFZpoolResult* item = ffListAdd(result);
+            ffStrbufInitS(&item->name, line);
+            ffStrbufSubstrBeforeFirstC(&item->name, '\t');
+            uint32_t startIndex = item->name.length;
+            ffStrbufInitS(&item->state, line + startIndex + sizeof("health\t"));
+            ffStrbufSubstrBeforeFirstC(&item->state, '\t');
+            ffStrbufGetline(&line, &len, &output);
+            item->version = ffStrbufToUInt(
+                &(FFstrbuf){.chars = line + startIndex + sizeof("version\t")} , 5000);
+            ffStrbufGetline(&line, &len, &output);
+            item->total = ffStrbufToUInt(
+                &(FFstrbuf){.chars = line + startIndex + sizeof("size\t")}, 0);
+            ffStrbufGetline(&line, &len, &output);
+            item->used = ffStrbufToUInt(
+                &(FFstrbuf){.chars = line + startIndex + sizeof("allocated\t")}, 0);
+            ffStrbufGetline(&line, &len, &output);
+            item->fragmentation = ffStrbufToDouble(
+                &(FFstrbuf){.chars = line + startIndex + sizeof("fragmentation\t")}, -DBL_MAX);
+        }
+        return NULL;
+    }
+    return "Failed to run zpool get -Hp health,version,size,allocated,fragmentation";
 }
 
 #endif


CarterLi:
Thank you for your contribution. I wonder why the zfs header is so different in NetBSD.

uzuki314:
Actually I'm not that familiar with NetBSD. I just tried multibooting Gentoo, FreeBSD, and NetBSD recently, and was surprised to find that NetBSD supports ZFS — so I managed to make them all boot from the same zpool.

As for your question, I tried to do some research: NetBSD’s ZFS implementation was originally ported from an older Solaris version (around the early OpenZFS days in 2009), and later rebased on FreeBSD’s code in 2018. This work was done before the OpenZFS and ZFS on Linux codebases were merged into OpenZFS 2.0 in 2020, so some parts of NetBSD’s implementation might still reflect the pre-merge structure. That might explain why some of the headers look a bit different. Because of that, it took me an afternoon to hunt down all the required types and definitions @@

reference:
https://en.wikipedia.org/wiki/NetBSD
https://wiki.netbsd.org/zfs/

CarterLi:
Your code adds a lot of complexity, which I don't really like. Can we just don't use dlopen just like what I did for SunOS?

uzuki314:
I followed your style and simplified the code this time.

> [Architectures](https://wiki.netbsd.org/zfs/#index5h2)
> ZFS is built by default on amd64 and aarch64 on netbsd-9. In 10 and current, it is also built by default on sparc64. Using it on other architectures requires a full release build with `MKZFS=yes`.

So while I set `ENABLE_LIBZFS` by default for NetBSD in CMakeLists, it might not be available on all architectures. 
Do you think we should still keep it enabled by default for NetBSD, or make it opt-in instead?

src/detection/zpool/libzfs_simplified_nbsd.h
```diff
@@ -0,0 +1,211 @@
...
+extern int zpool_iter(libzfs_handle_t *, zpool_iter_f, void *);
+extern uint64_t zpool_get_prop_int(zpool_handle_t *, zpool_prop_t, zprop_source_t *);
+extern zpool_status_t zpool_get_status(zpool_handle_t *, char **);
+extern nvlist_t *zpool_get_config(zpool_handle_t *, nvlist_t **);
+
+const char *zpool_get_state_str(zpool_handle_t *);
+
+// From https://github.com/NetBSD/src/blob/trunk/external/cddl/osnet/dist/uts/common/sys/nvpair.h
+
+extern nvlist_t *opensolaris_fnvlist_lookup_nvlist(nvlist_t *, const char *);
```
CarterLi: Does the following code work?

```c
extern nvlist_t *fnvlist_lookup_nvlist(nvlist_t *, const char *) __attribute__(("opensolaris_fnvlist_lookup_nvlist"));
```

src/detection/zpool/libzfs_simplified_nbsd.h
```diff
@@ -0,0 +1,211 @@
...
+extern int opensolaris_nvlist_lookup_uint64_array(nvlist_t *, const char *, uint64_t **, uint_t *);
+extern int opensolaris_nvlist_lookup_uint32_array(nvlist_t *, const char *, uint32_t **, uint_t *);
```
CarterLi: Can we merge this file into `libzfs_simplified.h`?

src/detection/zpool/zpool_linux.c
```diff
@@ -36,6 +36,37 @@ const char* zpool_get_state_str(zpool_handle_t* zpool)
...
+const char* zpool_get_state_str(zpool_handle_t* zpool)
+{
+    if (zpool_get_state(zpool) == POOL_STATE_UNAVAIL)
+        return "FAULTED";
+    else
+    {
+        const char *str;
+        zpool_status_t status = zpool_get_status(zpool, (char**) &str);
+        if (status == ZPOOL_STATUS_IO_FAILURE_WAIT ||
+            status == ZPOOL_STATUS_IO_FAILURE_CONTINUE)
```
CarterLi: Just use `#ifdef ZPOOL_STATUS_IO_FAILURE_MMP` in `__sun` part.

CarterLi:
> I followed your style and simplified the code this time.
> 
> > [Architectures](https://wiki.netbsd.org/zfs/#index5h2)
> > ZFS is built by default on amd64 and aarch64 on netbsd-9. In 10 and current, it is also built by default on sparc64. Using it on other architectures requires a full release build with `MKZFS=yes`.
> 
> So while I set `ENABLE_LIBZFS` by default for NetBSD in CMakeLists, it might not be available on all architectures. 
> Do you think we should still keep it enabled by default for NetBSD, or make it opt-in instead?

Better to check the existance of libzfs.so in CMake.

src/detection/zpool/libzfs_simplified_nbsd.h
uzuki314: Not work, maybe just use macro.

src/detection/zpool/libzfs_simplified_nbsd.h
uzuki314: Yes. Aside from zpool_prop_t (which differs), the rest can be shared.

src/detection/zpool/libzfs_simplified_nbsd.h
uzuki314: Or make an inline wrapper?

uzuki314:
Merge `libzfs_simplified_netbsd.h` into `libzfs_simplified.h`, add inline wrappers to translate the `opensolaris_*` symbols, and reuse `zpool_get_state_str`.

src/detection/zpool/libzfs_simplified.h
```diff
@@ -46,13 +56,16 @@ typedef enum {
     ZPOOL_PROP_FREE,
     ZPOOL_PROP_ALLOCATED,
     ZPOOL_PROP_READONLY,
+#ifndef __NetBSD__
     ZPOOL_PROP_ASHIFT,
```
CarterLi, line 49: This is an ABI break change and won't work in future zfs versions.

```
 * Pool properties are identified by these constants and must be added to the
 * end of this list to ensure that external consumers are not affected
 * by the change.
```

Bro...

src/detection/zpool/libzfs_simplified_nbsd.h
CarterLi: Sorry, it was a typo. It should be `__attribute__((alias("opensolaris_fnvlist_lookup_nvlist")));`

CarterLi:
I still don't like the change. I think the zfs code of NetBSD port is full of mess and I really don't want it to be embeded into fastfetch's repo. I would like just leave `item->state` empty.

CarterLi:
NetBSD itself is a niche system and NetBSD don't use zfs by default either. I don't really care about zfs support on NetBSD, to be honest.

uzuki314:
How about implementing it by just parsing the output of `zpool get -Hp version,size,free,fragmentation,health`?

CarterLi:
> How about implementing it by just parsing the output of `zpool get -Hp version,size,free,fragmentation,health`?

Just use the Command module then.

src/detection/zpool/libzfs_simplified.h
uzuki314: Bro... 🥺
The concern about ABI stability is valid — but the thing is, libzfs is not a stable interface in the first place.
As noted here: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=257222#c1

`libzfs` symbols are explicitly not guaranteed to be stable, the same zpool shows different fragmentation values on Gentoo and FreeBSD, and even within the same OS family (e.g., FreeBSD → TrueNAS), the ABI breaks frequently, including changes in internal struct layouts and symbol visibility.

So saying “this breaks ABI” is somewhat moot, because the whole module already relies on an ABI that upstream ZFS developers do not intend to keep stable.

Regarding NetBSD being niche — I agree it’s not a mainstream platform, but that's precisely why a simple, command-output-based fallback (zpool get -Hp ...) is a safer option than maintaining a separate set of libzfs headers for each OS.

If the goal is long-term maintainability, relying on zpool output (which is stable across Linux/FreeBSD/NetBSD/illumos(sunos)) is arguably more robust than relying on `libzfs` internals.

If full `libzfs` support is considered too fragile or high-maintenance, maybe it’s worth discussing whether the zpool module should:
* use libzfs only where it’s reliable, and
* fall back to parsing the CLI elsewhere, or
* rely entirely on CLI output for consistency across platforms.

I’m not advocating removing the module — ZFS users exist on all these platforms — but the current situation already depends on unstable APIs, and a CLI-based approach avoids exactly the ABI issues being pointed out.

CarterLi: > `libzfs` symbols are explicitly not guaranteed to be stable

I know ABI break change can be useful sometimes, but this breakage is completely useless and extremely dangerous. `ZPOOL_PROP_FRAGMENTATION` will be silently increased from 22 to 23, just because of they added a new entry at the MIDDLE of an enum. The change is ALWAYS wrong if they expose their enum outside the ABI boundry, NEVER makes sence and every C programmer should know it. They knew it actually (hence [the comments](https://github.com/NetBSD/src/blob/35a723947b61a9ad3992b2bb21e7a65ca8aa5046/external/cddl/osnet/dist/uts/common/sys/fs/zfs.h#L86-L87)). Your program won't crash, it will just query a wrong property and generate incorrect results.

uzuki314:
The reason to implement native parsing instead of using the Command module is that the Command module currently only supports single-line output.

Producing multiple lines is required when more than one pool exists. The Command module does not provide any mechanism to associate those lines with a single module output, so each line is printed as a separate field line. As a result, the output breaks alignment with the logo, and the Zpool entries appear interleaved inside the ASCII logo section instead of staying within the expected module column.

Because of this limitation, supporting multiple pools reliably requires parsing the output line by line inside the module itself, which is exactly what this implementation does.

This approach keeps the existing libzfs path untouched, but guarantees correct behavior everywhere else, including NetBSD, Windows, macOS, and other platforms where OpenZFS exists but the libzfs-based approach is not available.

If multi-pool support is required and the current Command module cannot provide it, then a parsing implementation is unavoidable. This is the smallest possible change that ensures correct behavior across all OpenZFS platforms without relying on unstable libzfs ABIs.

CarterLi:
Did my portable implementation. Please test it if you would like.

uzuki314:
that's sick

image

CarterLi:
Bro really use zfs on Windows...

Okey I just installed OpenZFS on Windows (at the risk of BSOD BTW), and found that OoW doesn't provide `libzfs.dll` or similar; only bunch of exes exist.

uzuki314:
BTW, macOS can almost work out of the box as long as libzfs.dylib is in the library path.

image

PR #2072, opened 2025-11-29
add kiss packages count

VicRadioHead:
Sorry to bother you, but this PR adds the kiss packages count support.
I knew the mainline of kiss linux and kiss package manager is dead. but it is still maintaining by unofficial community: https://kisscommunity.org/ , so I don't think it should be abandoned.
I promise I have tested this on my machine, its works fine.
I can't programming well, so I'm not sure if it can be used, I'm so sorry.
output

### DIFF
--- a/src/detection/packages/packages.h
+++ b/src/detection/packages/packages.h
@@ -21,6 +21,7 @@ typedef struct FFPackagesResult
     uint32_t guixUser;
     uint32_t hpkgSystem;
     uint32_t hpkgUser;
+    uint32_t kiss;
     uint32_t linglong;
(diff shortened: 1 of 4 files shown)


CarterLi:
LGTM.

---

Off topic. The ksh version detection seems broken. Does `ksh --version` work in your system?

VicRadioHead:
> LGTM.
> 
> Off topic. The ksh version detection seems broken. Does `ksh --version` work in your system?

No, OpenBSD's OKSH doesn't use this way to show version, but it can show version by output variable $KSH_VERSION

PR #2077, opened 2025-12-02
Implement Sonic Desktop (XLibre) branding detection

HaplessIdiot:
Add Sonic Desktop (XLibre) branding override check for new kwin-x11-sonic package on AUR https://aur.archlinux.org/packages/plasma-x11-session-sonic
https://aur.archlinux.org/packages/kwin-x11-sonic


CarterLi:
I don't think it even compiles because of the extra `}` bro...

PR #2095, opened 2025-12-15
docs(README): add mise alternative method installation

jylenhof:


CarterLi:
Sorry, but I don't like to add a neofetch installation method.

PR #2113, opened 2025-12-24
make nix detection faster

lhz07:
Counting nix-user and nix-default sequentially is quite slow.
On my MacBook, it takes around 120 ms (measured with `fastfetch --stat`).

This PR changes the implementation to call `nix-store` in parallel, reducing the time to about 80 ms on my machine.

It also removes the hash cache. In practice, the cache only improves performance by roughly 10 ms, while a cache miss (when the hash changes) can take up to 230 ms, making the overall behavior worse.

### DIFF
--- a/src/detection/packages/packages.h
+++ b/src/detection/packages/packages.h
@@ -1,10 +1,10 @@
 #pragma once
 
+#include "common/processing.h"
 #include "fastfetch.h"
 #include "modules/packages/option.h"
 
-typedef struct FFPackagesResult
-{
+typedef struct FFPackagesResult {
     uint32_t amSystem;
     uint32_t amUser;
     uint32_t apk;
@@ -47,7 +47,7 @@ typedef struct FFPackagesResult
     uint32_t winget;
     uint32_t xbps;
 
-    uint32_t all; //Make sure this goes last
+    uint32_t all; // Make sure this goes last
 
     FFstrbuf pacmanBranch;
 } FFPackagesResult;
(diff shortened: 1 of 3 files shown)


CarterLi:
The original author is @xoltia. @xoltia what's your opinion?

sabisho:
Yeah fastfetch on NixOS feels very slow at fetching info, hope this fixes that.

lhz07:
I don’t have Nix installed on my Linux machine, so this change only modifies the macOS-side package detection. But after a little modification, it should also work on NixOS or any Linux distribution where Nix is installed as a package manager.

However, the performance improvement is inherently limited. The main bottleneck is that spawning nix-store itself is expensive, taking roughly 70–80 ms per invocation. To reach performance comparable to other package managers, we need a more direct integration, for example:
- Using Nix’s C API (currently unstable, and it’s unclear whether it exposes the required functionality).
- Querying Nix’s internal SQLite database, which is not a public API and may change without notice.
- Communicating directly with the Nix daemon socket, which is also undocumented and unstable.

Or another way, we can make other parts faster. For example, connecting to the display server is also an expensive operation. On my MacBook, connecting to the display server takes about 40 ms. Combined with the ~80 ms cost of invoking nix-store, this adds up to roughly 120 ms in total.

Currently, fastfetch determines whether a feature needs to run while iterating over the JSON objects one by one. However, parsing the JSON into C structures is very cheap, usually taking around 1 ms or less. A more efficient approach would be to parse the JSON into C structures first, and then decide whether a thread needs to be spawned to connect to the display server. 

With this optimization, the display server connection can be performed in parallel with the Nix package query, so the 40 ms cost is effectively hidden within the nix-store execution time. As a result, the total time remains around 80 ms (excluding fastfetch’s own startup time).

PR #2116, opened 2025-12-26
the name it's not been used

Xray-OS:


CarterLi:
What's your opinion? @FabioLolix

FabioLolix:
Another rename this year but no rebranding to viper yet; let's wait

Xray-OS:
more fragmentation up the way

PR #2119, opened 2025-12-26
JsonSchema: correct references for keyboard, mouse items

3ventic:
- corrected format references from memoryFormat to keyboardFormat and mouseFormat respectively
- added outputColor to both since it appears to be supported for keyboard and mouse items too


3ventic:
Thank you Copilot for this anti-summary

CarterLi:
LGTM

PR #2123, opened 2025-12-29
Logo (Builtin): fix colors of deepin and UOS

qaqland:
The colors of both are in the blue range.

deepin

UOS

---

Small modification: when the word "deepin" appears alone or only combined with "linux", it is in lowercase.


CarterLi:
Thanks for your donation BTW :)

PR #2138, opened 2026-01-09
macOS large Apple logo: add missing warm band using ANSI LIGHT_RED

ctb248:
Hi! This PR improves the color accuracy of the large macOS Apple rainbow logo by adding the missing transition band between yellow and red.

The current logo renders consecutive ANSI_RED bands in the warm section, which compresses the  rainbow effect. This change inserts a LIGHT_RED (ANSI 91) band to (more closely) match the classic Apple rainbow mark. Sure everyone's seen it, but pasted for reference:
apple-small

Not sure if there's a technical/consistency rationale for not including the light/bright ANSI color here already, but the current color scheme immediately looked off when I ran fastfetch on a Macbook. Couldn't find any issues/discussions/previous PR's related to this, but could have missed something.

On terminals that distinguish normal vs. bright ANSI colors by default (notably Apple Terminal, where macOS is most likely), this produces a visibly improved and more accurate gradient.

### Notes

Affects only the large, default rainbow Apple logo rendered for macOS / OSX.

Just wanted to update the default logo, so no changes to logo variants/small versions. However, I did notice that the `macos3` variant may be misconfigured. It's using 6 positional color tokens, while only 5 values are defined in its color array, which results in an uncolored 6th band:

**Current macos3 appearance:**
Screenshot macos3

Change verified visually in:

Apple Terminal (distinct transition visible with default settings):
Screenshot Apple Terminal

VS Code integrated terminal (may collapse depending on theme/options. Disabled bold-as-bright for illustration):
**Original appearance:**
Screenshot 2026-01-09 at 4 49 46 PM

**New, bold-as-bright disabled:**
Screenshot 2026-01-09 at 4 49 31 PM

Also checked in iTerm, which renders bold-as-bright by default. In bold-as-bright/ANSI-only terminals, `BRIGHT_RED` will either collapse into `RED`, or `RED` will be rendered as `BRIGHT_RED` by default, resulting in 4 bands of the same color like today. So should be no regressions anywhere.

If this change is desired, happy to expand the PR or make a new one to add `LIGHT_RED` to other Apple logos, and correct the `macos3` variant rendering.

Thanks!


CarterLi:
LGTM, thanks.

PR #2157, opened 2026-01-18
Aritx logos improvements

thealio:
Hi 

proposal chenges

- Additional Artix big logo (see my issue)

-  The artix2_small.txt logo is cleary better , I think it should be the default for Artix linux small logo

## Checklist

- [x ] I have tested my changes locally.


CarterLi:
> Additional Artix big logo (see my issue)

It's ok.

> The artix2_small.txt logo is cleary better , I think it should be the default for Artix linux small logo

It's only your personal opinion. Let's not change the default before more people ask.

PR #2162, opened 2026-01-20
Improve HarmonyOS Detection Code

cyberMilosz:
## Summary

Current HarmonyOS detection code assumes that `uname` will accurately identify the OS. However, this is not the case on more recent emulators (as they seem to just use the Linux kernel):

```
# uname -a
Linux localhost 5.10.210 #1 SMP Tue Nov 11 15:26:56 CST 2025 x86_64
```

Real hardware still identifies itself correctly:

```
$ uname -a
HarmonyOS localhost HongMeng Kernel 1.10.5 #1 SMP Wed Dec 24 05:34:51 UTC 2025 aarch64
```

Instead of `uname`, we should rely on `param` output - we can get the OS name, version, API version, and security patch like so:

Emulator:
```
# param get const.product.os.dist.name
HarmonyOS
# param get const.product.os.dist.version
6.0.0
# param get const.product.incremental.version
6.0.1.112
# param get const.ohos.apiversion
21
# param get const.product.software.version.name
HarmonyOS NEXT Developer Beta1
# param get const.product.software.version
emulator 6.0.0.112(SP3DEVC00E112R4P11)
# param get const.ohos.version.security_patch
2025/11/01
```

Real hardware (note `param get const.product.software.version.name` appears to be a recent addition, so it's missing here):
```
$ param get const.product.os.dist.name
HarmonyOS
$ param get const.product.os.dist.version
5.1.0
$ param get const.product.incremental.version
5.1.1.208
$ param get const.ohos.apiversion
19
$ param get const.product.software.version.name
Get parameter "const.product.software.version.name" fail! errNum is:106!
$ param get const.product.software.version
TXZ-W10M 5.1.0.151(C00E151R4P1)
$ param get const.ohos.version.security_patch
2026/01/01
```

Tested on multiple Huawei emulators [6.0.1(21), 5.1.1(19), 5.1.0(18)] and a Huawei MatePad 11.5 running HarmonyOS 5.1.0.151

Example output from a current emulator:

image
 

## Changes

- Introduce HarmonyOS detection logic based on `param` output, falling back to existing `uname` logic as a last resort.

## Checklist

### DIFF
--- a/src/detection/os/os_linux.c
+++ b/src/detection/os/os_linux.c
@@ -330,6 +330,140 @@ static bool detectBedrock(FFOSResult* os)
     return false;
 }
 
+static bool detectHarmonyOS(FFOSResult* os)
+{
+    // HarmonyOS 5+ uses param command to read system parameters
+    // Check for const.product.os.dist.name=HarmonyOS
+    FF_STRBUF_AUTO_DESTROY osName = ffStrbufCreate();
+    if (ffProcessAppendStdOut(&osName, (char* const[]) {
+        "param", "get", "const.product.os.dist.name",
+        NULL,
+    }) != NULL)
+        return false;
+
+    ffStrbufTrimRightSpace(&osName);
+    if (!ffStrbufEqualS(&osName, "HarmonyOS"))
+        return false;
+
+    // Detected HarmonyOS - read version and other info
+    ffStrbufSetS(&os->id, "harmonyos");
+    ffStrbufSetS(&os->idLike, "harmonyos");
+    ffStrbufSetS(&os->name, "HarmonyOS");
+
+    // Read base version (e.g., 6.0.0)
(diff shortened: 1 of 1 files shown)


CarterLi:
I assume the HarmonyOS command `param` is a variant of the Android command `getprop`. Is there an API like `__system_property_get` in Android? You spawn 8 processes in one module which is not acceptable to me.

In addition, please move HarmonyOS detection code after the standard `os-release` code, so it don't affect performance of major Linux distros.

cyberMilosz:
> I assume the HarmonyOS command `param` is a variant of the Android command `getprop`.

Yeah

> Is there an API like `__system_property_get` in Android?

Looks like `GetParameter()` from https://gitee.com/openharmony/startup_init/blob/master/interfaces/innerkits/syspara/parameter.c is the right answer here. I'll play around with it when I next have time and try to suggest a cleaner implementation.

PR #2180, opened 2026-02-12
feat: add 6-color support to nixos logo

klercke:
## Summary

This PR adds a color selector to each lambda in the NixOS logo, allowing them to each be colored individually. It does not change the default output:

image

image

N/A - Modifies an existing logo

## Changes

- Sets 6 color options in src/logo/ascii/nixos.txt
- Defines the 6 colors in src/logo/builtin.c such that the logo is displayed as it was previously by default

## Checklist


wasdhjklxyz:
Bruh why not make a nixos_multicolor* variant instead of changing how the coloring works

My config was written for two colors and this fucks it up... and what if i preferred the two colors

imo should guarantee that old configs work across same major versions (was on v2.43.0 and this released in v2.59.0!!!!?)

Related: #2357

klercke:
> Bruh why not make a nixos_multicolor* variant instead of changing how the coloring works

I'm sorry that you don't agree with the change, but my reasoning behind adjusting the default logo was that the [official NixOS Branding Guide](https://nixos.org/branding/) include a multi-color version (see pages 21 and 22 of the PDF linked on that page).

I think that ultimately having to define a few more colors is more intuitive for end users than having to look at documentation to determine that there are multiple variants of the logo. This was my experience: I was confused as to why the logo only supported two colors when NixOS uses 6 in their own branding.

wasdhjklxyz:
> ...the [official NixOS Branding Guide](https://nixos.org/branding/) include a multi-color version (see pages 21 and 22 of the PDF linked on that page).

according to that, then you shouldve made a new file named `nixos_rainbow.txt` and left `nixos.txt` alone

also you couldve at least updated the colors to actually be the rainbow variant instead of just using the blue and aqua. you added 4 more color slots just to repeat blue & aqua not actually changing anything other than the amount of colors to process just to have the same end result (in default configs)

> I think that ultimately having to define a few more colors is more intuitive for end users than having to look at documentation to determine that there are multiple variants of the logo.

after this pr was merged, everyone that used colors other than the default had to update their config and it wasnt as intuitive as you think. there was no error or warning (which is a problem itself) that i was missing 4 more colors. ihad to look at a diff to figure out 4 more colors were added

you dont even have to look @ the docs. youd see the rainbow variant right next to the other nix ones when using `--list-logos`

> I was confused as to why the logo only supported two colors when NixOS uses 6 in their own branding.

they use the default blue and aqua one literally everywhere. the only place i remember seeing the rainbow variant was in the downloads page i think. which is why i think you kept the blue and aqua instead of updating to actually use all 6 colors

CarterLi:
Even if OP adds a logo called nixos_rainbow at the beginning, I will ask him to update the original one.
I knew the change might break some configs, but it wasnt a blocking issue. There are Discussion and Issue sections in Github to asking for help, and I am still active (as for now) to answer questions. The fix is simple too.
Calm down bro.

PR #2184, opened 2026-02-14
Global (macOS): enable X11 related stuff

robertkirkman:
## Summary

https://github.com/fastfetch-cli/fastfetch/commit/7d92da88294cc3f71dbe1e5dc9203f331a42836e but for macOS

## Changes

- Enable on macOS the detection of:
  - X11
  - DBus
  - PulseAudio

- Enable printing window manager protocol name on macOS when X11 is not running as "CoreGraphics"

## Checklist


robertkirkman:
image

CarterLi:
I don't like to add code complexity for features that no one uses. Sorry.

CarterLi:
@robertkirkman Do you really use X11 in macOS, or just install it for testing?

Do you use macOS?

robertkirkman:
> @robertkirkman Do you really use X11 in macOS, or just install it for testing?
> 
> Do you use macOS?

Actually, you are right, I do not frequently use macOS, only for testing and development purpose, and I just created this PR because of when I said on the 7th July 2025 that I was going to create a PR to add X11 to fastfetch for macOS and Android but then you made it for Android so now it's only macOS

- https://github.com/termux/termux-packages/pull/25266#issuecomment-3046669202


When I use macOS I use this macOS QEMU/KVM virtual machine to run it

https://github.com/kholia/OSX-KVM

I do understand that probably very few people use or want to use X11 in macOS so it makes sense that this is too much added complexity for not enough users.

PR #2200, opened 2026-02-24
Add Zed support for editor detection

Alxhr0:
## Summary

Zed support for the editor module

## Changes

- Zed detection in the editor module

## Checklist

### DIFF
--- a/src/detection/editor/editor.c
+++ b/src/detection/editor/editor.c
@@ -115,7 +115,8 @@ const char* ffDetectEditor(FFEditorResult* result)
         ffStrbufEqualS(&result->exe, "hx") ||
         ffStrbufEqualS(&result->exe, "code") ||
         ffStrbufEqualS(&result->exe, "pluma") ||
-        ffStrbufEqualS(&result->exe, "sublime_text")
+        ffStrbufEqualS(&result->exe, "sublime_text") ||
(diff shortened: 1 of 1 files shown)


src/detection/editor/editor.c
```diff
@@ -115,7 +115,8 @@ const char* ffDetectEditor(FFEditorResult* result)
         ffStrbufEqualS(&result->exe, "hx") ||
         ffStrbufEqualS(&result->exe, "code") ||
         ffStrbufEqualS(&result->exe, "pluma") ||
-        ffStrbufEqualS(&result->exe, "sublime_text")
+        ffStrbufEqualS(&result->exe, "sublime_text") ||
+        ffStrbufEqualS(&result->exe, "zeditor")
```
CarterLi, line 119: Why is it called `zeditor` instead of `zed`?

Alxhr0: That's the binary name in the package on Arch, seems to also be the case on NixOS

Maybe some distro's do call it zed-editor, but I haven't seen one do that yet

CarterLi: On macOS, it's just called `zed`. Hence the question.

Alxhr0: Ah, I see

I'm at school so I can't really add "zed" to the code, sorry about that

PR #2220, opened 2026-03-10
Keyboard (Linux) Bluetooth support

### LINKED ISSUE #2219: [BUG] Keyboard module doesn't detect Bluetooth keyboards on Linux
## Bug description

The Keyboard module returns empty results for Bluetooth-connected keyboards on Linux. The kernel correctly recognizes the device, but fastfetch doesn't find it.

## Fastfetch version

`fastfetch 2.58.0 (x86_64)`

## OS / Environment

Linux 6.19.6-gentoo-dist (x86_64)

## Steps to reproduce

1. Connect a keyboard via Bluetooth (in my case a ZMK-based split keyboard)
2. Run `fastfetch -s keyboard --logo none --format json`
3. Result: `"result": []`

## Diagnostic details

The kernel sees the keyboard correctly in `/proc/bus/input/devices`:

```
I: Bus=0005 Vendor=1d50 Product=615e Version=0001
N: Name="Toucan Keyboard"
P: Phys=a0:ad:9f:72:f6:6b
S: Sysfs=/devices/virtual/misc/uhid/0005:1D50:615E.0019/input/input35
U: Uniq=f4:ea:30:70:68:4e
H: Handlers=sysrq kbd event3
```

It has the `kbd` handler assigned, confirming it is a keyboard.

However, because it connects via Bluetooth (bus `0005`, via `uhid`), udev does not create a symlink in `/dev/input/by-id/`. It appears fastfetch's keyboard detection on Linux relies on enumerating `/dev/input/by-id/`, so Bluetooth keyboards are missed entirely.

## Expected behavior

Bluetooth keyboards with a `kbd` handler should be detected. A fallback to `/proc/bus/input/devices` or `/sys/class/input/` would cover devices that don't have a `/dev/input/by-id/` entry.

## Workaround

Adding a udev rule to manually create a `by-id` symlink makes fastfetch detect the keyboard:

```
# /etc/udev/rules.d/99-toucan-keyboard.rules
SUBSYSTEM=="input", ATTRS{name}=="Toucan Keyboard", SYMLINK+="input/by-id/bluetooth-Toucan_Keyboard-event-kbd"
```

bennypowers:
## Summary

Adds bluetooth keyboard support. This adds a fallback that parses `/proc/bus/input/devices` for devices with a `kbd` handler, filtered by `EV_REP` capability to exclude pseudo-keyboards like "Power Button".

Closes #2219 

## Changes

- Extracted shared sysfs name/serial reading into `addDevice()` helper
- Moved the existing by-path scan into `detectFromByPath()`, now returns `bool`
- Added `detectFromProc()` — parses `/proc/bus/input/devices` for `kbd` handler entries
- Added `hasAutoRepeat()` — checks `EV_REP` capability in sysfs to filter pseudo-keyboards
- `ffDetectKeyboard()` runs both discovery paths; dedup bitmap prevents double-counting; returns error only if both fail

## Checklist

## Alternatives considered:
- Replace by-path entirely with proc — simpler single path, but changes behavior for all existing users
- Scan `/sys/class/input/eventN/` directly and check capabilities — avoids proc text parsing but requires more sysfs reads per device
- Filter by `sysrq` handler instead of `EV_REP` — no extra sysfs read needed, but would miss compact keyboards without a SysRq key
- Filter by bus type (USB=3, BT=5, I2C=0x18) — very precise but fragile allowlist to maintain
- No filter, rely on `ignores` config — simplest code but worse defaults

### DIFF
--- a/src/detection/keyboard/keyboard_linux.c
+++ b/src/detection/keyboard/keyboard_linux.c
@@ -48,66 +48,56 @@ const char* ffDetectKeyboard(FFlist* devices /* List of FFKeyboardDevice */)
 
     uint64_t flags = 0;
     FF_STRBUF_AUTO_DESTROY path = ffStrbufCreate();
+    FFstrbuf kbd = ffStrbufCreateStatic("kbd");
 
-    const char* line = content.chars;
-    while (line && *line)
+    char* line = NULL;
+    size_t len = 0;
+    while (ffStrbufGetline(&line, &len, &content))
     {
-        if (ffStrStartsWith(line, "H: Handlers="))
+        if (!ffStrStartsWith(line, "H: Handlers="))
+            continue;
+
+        const char* handlers = line + strlen("H: Handlers=");
+
+        if (!ffStrbufMatchSeparatedS(&kbd, handlers, ' '))
+            continue;
+
+        // Find "eventN" token and extract the index
+        const char* eventStr = strstr(handlers, "event");
+        if (!eventStr)
+            continue;
+
+        char* pend = NULL;
+        uint32_t eventIndex = (uint32_t) strtoul(eventStr + strlen("event"), &pend, 10);
+        if (pend == eventStr + strlen("event"))
+            continue;
+
+        // Skip duplicates (dedup bitmap covers indices 0-63; higher indices are not deduped)
+        if (eventIndex < 64 && (flags & (1ULL << eventIndex)))
+            continue;
+
+        if (!isRealKeyboard(eventIndex, &path))
+            continue;
+
+        ffStrbufSetF(&path, "/sys/class/input/event%u/device/name", (unsigned) eventIndex);
+
+        FF_STRBUF_AUTO_DESTROY name = ffStrbufCreate();
+        if (ffAppendFileBuffer(path.chars, &name))
         {
-            const char* handlers = line + strlen("H: Handlers=");
-            bool hasKbd = false;
-            uint32_t eventIndex = UINT32_MAX;
+            if (eventIndex < 64)
+                flags |= (1ULL << eventIndex);
 
-            // Parse space-separated handler names
-            const char* p = handlers;
-            while (*p && *p != '\n')
-            {
-                while (*p == ' ') p++;
-                if (*p == '\n' || *p == '\0') break;
+            ffStrbufTrimRightSpace(&name);
+            ffStrbufSubstrBefore(&path, path.length - (uint32_t) strlen("name"));
 
-                const char* wordStart = p;
-                while (*p && *p != ' ' && *p != '\n') p++;
-                uint32_t wordLen = (uint32_t)(p - wordStart);
+            FFKeyboardDevice* device = (FFKeyboardDevice*) ffListAdd(devices);
+            ffStrbufInitMove(&device->name, &name);
+            ffStrbufInit(&device->serial);
 
-                if (wordLen == 3 && memcmp(wordStart, "kbd", 3) == 0)
-                    hasKbd = true;
-                else if (wordLen > strlen("event") && memcmp(wordStart, "event", strlen("event")) == 0)
-                {
-                    char* pend = NULL;
-                    eventIndex = (uint32_t) strtoul(wordStart + strlen("event"), &pend, 10);
-                    if (pend == wordStart + strlen("event")) eventIndex = UINT32_MAX;
-                }
-            }
-
-            // Skip duplicates (dedup bitmap covers indices 0-63; higher indices are not deduped)
-            bool seen = eventIndex < 64 && (flags & (1ULL << eventIndex));
-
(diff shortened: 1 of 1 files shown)


src/detection/keyboard/keyboard_linux.c
```diff
@@ -2,16 +2,35 @@
...
-const char* ffDetectKeyboard(FFlist* devices /* List of FFKeyboardDevice */)
+static void addDevice(FFlist* devices, uint64_t* flags, uint32_t index, FFstrbuf* path)
 {
-    // There is no /sys/class/input/kbd* on Linux
-    FF_AUTO_CLOSE_DIR DIR* dirp = opendir("/dev/input/by-path/");
-    if (dirp == NULL)
-        return "opendir(\"/dev/input/by-path/\") == NULL";
+    if (index >= 64 || (*flags & (1UL << index)))
+        return;
+    *flags |= (1UL << index);
```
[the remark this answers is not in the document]
bennypowers: how likely is it for users to have 64 input devices?

CarterLi:
What about removing `/dev/input/by-path/` completely, so the detection also work if both wired and bluetooth keyboards are connected (should be possible on laptops)?

bennypowers:
How's that look? It has to filter by KEY_A to avoid pulling in things like Bluetooth headphones - this will exclude macropads, but maybe that could be a later feature

src/detection/keyboard/keyboard_linux.c
```diff
@@ -2,55 +2,112 @@
...
+            const char* handlers = line + strlen("H: Handlers=");
+            bool hasKbd = false;
+            uint32_t eventIndex = UINT32_MAX;
 
-        char buffer[32]; // `../eventXX`
-        ssize_t len = readlinkat(dirfd(dirp), entry->d_name, buffer, ARRAY_SIZE(buffer) - 1);
-        if (len <= (ssize_t) strlen("../event") || !ffStrStartsWith(buffer, "../event")) continue;
-        buffer[len] = 0;
+            // Parse space-separated handler names
+            const char* p = handlers;
```
CarterLi: There's `ffStrbufGetline`

src/detection/keyboard/keyboard_linux.c
```diff
@@ -2,55 +2,112 @@
...
 
-        const char* eventid = buffer + strlen("../event");
+                const char* wordStart = p;
+                while (*p && *p != ' ' && *p != '\n') p++;
+                uint32_t wordLen = (uint32_t)(p - wordStart);
 
-        char* pend = NULL;
-        uint32_t index = (uint32_t) strtoul(eventid, &pend, 10);
-        if (pend == eventid) continue;
+                if (wordLen == 3 && memcmp(wordStart, "kbd", 3) == 0)
```
CarterLi: You may use `ffStrbufMatchSeparatedS`. Something like:

```c
FFstrbuf kbd = ffStrbufInitStatic("kbd");
ffStrbufMatchSeparatedS(&kbd, "sysrq kbd leds event3", ' '); // true
```

CarterLi:
> Bluetooth headphones

Bluetooth headphones also marked as kbd?

bennypowers:
yep, it picked up my sennheisers, which have inputs like next track play pause etc

src/detection/keyboard/keyboard_linux.c
bennypowers: Thanks, PTAL

src/detection/keyboard/keyboard_linux.c
```diff
@@ -2,44 +2,91 @@
...
+        if (!ffStrStartsWith(line, "H: Handlers="))
             continue;
 
-        char buffer[32]; // `../eventXX`
-        ssize_t len = readlinkat(dirfd(dirp), entry->d_name, buffer, ARRAY_SIZE(buffer) - 1);
-        if (len <= (ssize_t) strlen("../event") || !ffStrStartsWith(buffer, "../event")) continue;
-        buffer[len] = 0;
+        const char* handlers = line + strlen("H: Handlers=");
+
+        if (!ffStrbufMatchSeparatedS(&kbd, handlers, ' '))
```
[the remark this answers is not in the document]
bennypowers: this comment suggests that it would be better not to use the helpers. how would you like me to proceed? ignore the comment? revert? implement the suggestion?

CarterLi:
@bennypowers Please test [4a4afda](https://github.com/fastfetch-cli/fastfetch/commit/4a4afdadb802311cd965e15a088a843663d92c11)

bennypowers:
LGTM. Thanks for merging

```
Keyboard: Toucan Keyboard
Mouse 1: Toucan Mouse
Mouse 2: Logitech MX Ergo Multi-Device Trackball
```

JoeHelbing:
Gracias for this. Just switched from Windows to Linux and was wondering why the Toucan BT wasn't playing nice.

PR #2231, opened 2026-03-17
Differentiate @world and non-@world sets in package field

### LINKED ISSUE #2230: [FEAT] Differentiate emerge package counts by set (@world, @system, deps)
### Description

General inquiry and asking for guidance on implementing set differentiation for Portage sets.

### Motivation

Currently, fastfetch reports all emerge packages as a single count, conflating `@world`, `@system`, and pulled-in dependencies. Poor Portage getting discriminated while nixpkgs gets first-class treatment..

As per the [Gentoo Wiki on Package Sets](https://wiki.gentoo.org/wiki/Package_sets), `@system` and `@profile` are *"defined by the Gentoo development team"* as the base set of packages required for the OS to function, so essentially system-level metadata.

Since fastfetch already differentiates `nix-system` and `nix-user`, Portage deserves the same parity. Calculating these essentially costs nothing, it's just reading `/var/lib/portage/world` and the profile's `packages` file.

Counts would be mutually exclusive: `total - (@world + @system) = deps`. Two formats on the table:

**1:**
```
packages: 76 (@world), 50 (@system), 919 (emerge-deps), 12 (flatpak)
```
**2** (much like `nix-system`/`nix-user`):
```
packages: 76 (emerge-world), 50 (emerge-system), 919 (emerge-deps), 12 (flatpak)
```
or whichever you guys prefer, I'll try and see.

### Additional context

_No response_

impietic:
resolves #2230 and #2229 (gonna link them here anyway even if they're closed)

image
basically implements set differentiation for emerge pkgs, splitting the single <code>emerge</code> count into <code>emerge-world</code>, and <code>emerge-system</code>.
<br /><br />
detection reads <code>/var/lib/portage/world</code> for <code>@world</code>, and non-<code>@world</code> (a.k.a. <code>emerge-system</code>), and derives non-<code>@world</code> via <code>total - world</code>.
<br/><br/>
For Copilot:<br/>
1. <code>ffStrbufCountC(content, '\n')</code> computation is intentional  since this is a fetch tool, not a dependency resolver<br/>
2. <code>@system</code> set pkg counting was previously present, but removed as per maintainer's advice because the code underneath is really fragile<br/>
3. non-<code>@world</code> pkgs and dependencies were consolidated to <code>emerge-system</code> because in Gentoo, you technically can't remove either since there'll be dependents, much like a system pkg itself. Who even pulls in Glibc intentionally?


impietic:
thanks, on it

src/detection/packages/packages.h
```diff
@@ -48,7 +50,8 @@ typedef struct FFPackagesResult
     uint32_t winget;
     uint32_t xbps;
 
-    uint32_t all; //Make sure this goes last
+    uint32_t all; // Make sure this goes last among package counter fields
+    uint32_t emerge;
```
CarterLi: This is unacceptable. It must be calcuated with `emergeDeps+emergeSys+emergeWorld`

src/modules/packages/packages.c
```diff
@@ -473,7 +489,10 @@ FFModuleBaseInfo ffPackagesModuleInfo = {
         {"Pacman branch on manjaro", "pacman-branch"},
         {"Number of dpkg packages", "dpkg"},
         {"Number of rpm packages", "rpm"},
-        {"Number of emerge packages", "emerge"},
+        {"Number of all combined emerge packages", "emerge"},
```
CarterLi: Please follow the pattern of `*-all`, and move it to the bottom of the list

src/modules/packages/packages.c
```diff
@@ -159,7 +169,10 @@ bool ffPrintPackages(FFPackagesOptions* options)
             FF_ARG(counts.pacmanBranch, "pacman-branch"),
             FF_ARG(counts.dpkg, "dpkg"),
             FF_ARG(counts.rpm, "rpm"),
-            FF_ARG(counts.emerge, "emerge"),
+            FF_ARG(emergeAll, "emerge"),
```
CarterLi: Should be `emerge-all`

src/detection/packages/packages_linux.c
```diff
@@ -451,12 +452,179 @@ static uint32_t getPacmanPackages(FFstrbuf* baseDir)
     return getNumElements(baseDir, dbPath.chars, true);
 }
 
+static uint32_t getProfSysPackages(FFstrbuf* profileDir, uint32_t depth)
+{
+    if (depth > 64)
```
CarterLi: Why is it so deep?

src/modules/packages/packages.c
impietic: noted

src/detection/packages/packages_linux.c
impietic: left that as flat base, ill go lower it to around 16 or smth

CarterLi:
~What on earth does `/var/lib/portage/world` look like? Can you upload a sample file?~

I checked the file in <https://distrosea.com>. I don't see a comment line. If it's a auto-generated file, I don't think it will contain comment or empty line.

impietic:
> What on earth does `/var/lib/portage/world` look like? Can you upload a sample file?
```bash
➜  anthony@asta ~ cat /var/lib/portage/world
app-admin/doas
app-arch/7zip
app-arch/unrar
app-containers/podman
app-editors/nano
app-editors/vim
app-eselect/eselect-repository
app-misc/fastfetch
app-misc/jq
app-misc/sl
app-portage/eix
app-portage/genlop
app-portage/gentoolkit
app-shells/nushell
app-text/ghostscript-gpl
app-text/mupdf
dev-dotnet/dotnet-sdk-bin
dev-lang/zig
dev-util/android-tools
dev-util/pkgcheck
dev-util/pkgdev
games-misc/cowsay
games-misc/nyancat
gnome-base/dconf-editor
gnome-base/gnome-light
gnome-extra/gnome-browser-connector
gnome-extra/gnome-shell-extensions
gnome-extra/gnome-tweaks
llvm-core/lld
media-fonts/noto
media-fonts/noto-cjk
media-gfx/imagemagick
media-gfx/iscan
media-gfx/sane-backends
media-libs/libsdl2
media-libs/libv4l
media-libs/tiff
media-video/celluloid
media-video/ffmpeg
net-dns/bind-tools
net-misc/networkmanager
net-misc/whois
net-misc/yt-dlp
net-p2p/transmission
net-print/cups
net-print/cups-filters
net-print/cups-meta
net-print/epson-inkjet-printer-escpr
net-print/libcupsfilters
net-wireless/wireless-regdb
sys-apps/baobab
sys-apps/cpuid
sys-apps/mlocate
sys-apps/pciutils
sys-apps/qdiskinfo
sys-apps/ripgrep
sys-apps/systemd
sys-auth/rtkit
sys-block/io-scheduler-udev-rules
sys-boot/efibootmgr
sys-boot/gnu-efi
sys-boot/grub
sys-boot/plymouth
sys-firmware/intel-microcode
sys-firmware/sof-firmware
sys-fs/btrfs-progs
sys-fs/dosfstools
sys-fs/ntfs3g
sys-kernel/git-sources
sys-kernel/installkernel
sys-kernel/linux-firmware
sys-kernel/linux-headers
sys-libs/glibc
sys-power/thermald
www-client/brave-browser
www-client/lynx
x11-terms/ptyxis
```

it is basically a file that lists out the packages that I explicitly added through emerge.

CarterLi:
@hydrospherics Why do you think it may contain empty or comment lines?

impietic:
> @hydrospherics Why do you think it may contain empty or comment lines?

it usually doesnt (portage handles it fine), but copilot suggested that I harden the parsing anyway.

maybe I guess.. some people edit their world file manually?

CarterLi:
People can even run `rm -rf --no-preserve-root /`. I don't think fastfetch wants to support the case.

Counting the number of `\n` of the file should be ok.

src/detection/packages/packages_linux.c
```diff
@@ -451,12 +452,179 @@ static uint32_t getPacmanPackages(FFstrbuf* baseDir)
...
+
+                /* Count only non-empty, non-comment lines */
+                if (p < eol && *p != '#')
+                    ++world;
+
+                line = (*eol) ? eol + 1 : eol;
+            }
+        }
+        ffStrbufSubstrBefore(baseDir, baseDirLen);
+    }
```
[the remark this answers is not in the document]
impietic: This is just plain wrong. world file is ONLY used for individual package atoms like `app-editors/vim`. It cannot contain @/sets.

src/detection/packages/packages_linux.c
```diff
@@ -451,12 +452,179 @@ static uint32_t getPacmanPackages(FFstrbuf* baseDir)
...
+                ffStrbufSet(&parentPath, profileDir);
+                ffStrbufAppendNS(&parentPath, lineLen, line);
+            }
+
+            char resolved[PATH_MAX] = {0};
+            if (realpath(parentPath.chars, resolved))
+            {
+                FF_STRBUF_AUTO_DESTROY resolvedBuf = ffStrbufCreateS(resolved);
+                count += getProfSysPackages(&resolvedBuf, depth + 1);
+            }
```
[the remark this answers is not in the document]
impietic: too much for just a fetch tool

CarterLi:
The code of counting `@system` packages seems over complicated and vulnerable to me. Can we just counting `@world` and `non-@world`?

impietic:
> People can even run `rm -rf --no-preserve-root /`. I don't think fastfetch wants to support the case.
> 
> Counting the number of `\n` of the file should be ok.

true, I do that alot for some reason.. but AI is AI, it wants to handhold in some way.

impietic:
> The code of counting `@system` packages seems over complicated and vulnerable to me. Can we just counting `@world` and `non-@world`?

I see, so I guess I'll go rm counting the `@system` set and just resort to smth like:
`emerge-world`, `emerge-non-world` (`/var/db/pkg/*/*` count minus `@world` count), and `emerge-all` for all combined

hell, maybe even count non-world as `emerge-system` instead since you can't really remove them unless you unmerge forcefully.

src/modules/packages/packages.c
```diff
@@ -440,6 +452,7 @@ bool ffGeneratePackagesJsonResult(FF_MAYBE_UNUSED FFPackagesOptions* options, yy
     FF_APPEND_PACKAGE_COUNT(sorcery)
     FF_APPEND_PACKAGE_COUNT(winget)
     FF_APPEND_PACKAGE_COUNT(xbps)
+    yyjson_mut_obj_add_uint(doc, obj, "emerge-all", counts.emergeWorld + counts.emergeSystem);
```
CarterLi: Please follow the pattern above, and print world and system independently.

CarterLi (changes requested):

CarterLi:
@ceamac What's your opinion?

TLDR: This PR splits emerge packages count into `emerge-world` and `emerge-system` (actually non-world, i.e. `@all` - `@world`). It used to fetch `@system` and `@dep` packages count but was removed due to the complexity and vulnerability of the code.

src/modules/packages/packages.c
```diff
@@ -473,7 +485,8 @@ FFModuleBaseInfo ffPackagesModuleInfo = {
         {"Pacman branch on manjaro", "pacman-branch"},
         {"Number of dpkg packages", "dpkg"},
         {"Number of rpm packages", "rpm"},
-        {"Number of emerge packages", "emerge"},
+        {"Number of @world emerge set packages", "emerge-world"},
+        {"Number of non-world emerge set packages", "emerge-system"},
```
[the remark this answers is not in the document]
impietic: no longer exclusive to @system

src/modules/packages/packages.c
impietic: my bad, I was already asleep at that time

thanks for taking care of it though,  I appreciate it :)

impietic:
force pushed to amend the commit message from the merge new changes thing (I did that on web).

Also, we could consider aliasing `emerge` to `emerge-all` so we don't break the current configs people have that use `emerge`.

CarterLi:
LGTM, but I don't know if `emerge-system` causes confusion to other gentoo users.

ceamac:
> @ceamac What's your opinion?
> 
> TLDR: This PR splits emerge packages count into `emerge-world` and `emerge-system` (actually non-world, i.e. `@all` - `@world`). It used to fetch `@system` and `@dep` packages count but was removed due to the complexity and vulnerability of the code.

Yes, `@system` is particularly hard to count because it depends on the current profile, profiles can include other profiles, and users can even use custom profiles, it would not be straight forward to determine which packages it contains.

Most distributions keep track of which packages are manually installed and which are dependencies only, so they can be cleaned when no longer needed (see `apt-mark` in Debian or `pacman --asdeps` in Arch and so on).  I don't know if this is a metric that should be added to some or all distributions, that is not for me to decide.  In Gentoo this count is not too easy to get accurately.

Other than the `@world` set we also have custom sets, which also contain packages that the user has installed explictly, so they will not be autocleaned.  You can find them in `/var/lib/portage/world_sets` with the contents in `etc/portage/sets`.  For example in my case (I assume this is a popular one)

```
$ cat /var/lib/portage/world_sets
@esteam
```

Where `/etc/portage/sets/esteam` has the same format as the world file, but being in `/etc` it is expected to be edited manually, so it can contain comments and empty lines (`esteam` in particular is managed by `games-util/esteam` from the `steam` overlay, but it still containes comments).

Some people have more manually defined sets, so counting only the packages from the `world` file does not return the whole picture.  I would say the world file is a pretty good approximation, it covers most uses, but not all.

I will look over the implementation in the weekend, now I only read the discussion.

impietic:
> Yes, `@system` is particularly hard to count because it depends on the current profile, profiles can include other profiles, and users can even use custom profiles, it would not be straight forward to determine which packages it contains.
> 
> Most distributions keep track of which packages are manually installed and which are dependencies only, so they can be cleaned when no longer needed (see `apt-mark` in Debian or `pacman --asdeps` in Arch and so on).  I don't know if this is a metric that should be added to some or all distributions, that is not for me to decide.  In Gentoo this count is not too easy to get accurately.
> 
> Other than the `@world` set we also have custom sets, which also contain packages that the user has installed explictly, so they will not be autocleaned.  You can find them in `/var/lib/portage/world_sets` with the contents in `etc/portage/sets`.  For example in my case (I assume this is a popular one)
> 
> ```
> $ cat /var/lib/portage/world_sets
> @esteam
> ```
> 
> Where `/etc/portage/sets/esteam` has the same format as the world file, but being in `/etc` it is expected to be edited manually, so it can contain comments and empty lines (`esteam` in particular is managed by `games-util/esteam` from the `steam` overlay, but it still containes comments).
> 
> Some people have more manually defined sets, so counting only the packages from the `world` file does not return the whole picture.  I would say the world file is a pretty good approximation, it covers most uses, but not all.
> 
> I will look over the implementation in the weekend, now I only read the discussion.

Yeah, looking back I have no idea what I was thinking trying to count `@system`,  I was using plain asterisks to count the packages there, which was.. vulnerable, to say.

As for `world_sets`, I'd suggest counting them under emerge-world if we ever get to it. Merging them into a separate counter would just pollute the display for users with a lot of custom sets, and semantically they're still explicitly installed packages anyway.

Anyways, thanks for dropping by for now though.

impietic:
> LGTM, but I don't know if `emerge-system` causes confusion to other gentoo users.

Ahh, I didn't mean that it would cause confusion, it's just that, some people might not have their configs up ahead so yea.

Your call though.

Issue #1995, opened 2025-10-03
[FEAT] Subtract ZFS ARC from memory usage


leo60228:
### Description

The displayed memory in use should exclude ZFS's ARC, similarly to tools like htop.

### Motivation

Implementing this in a shell is non-trivial, and it's a logical extension of existing functionality.

### Additional context

Due to Linux kernel limitations, ZFS's in-memory caches aren't included in `MemAvailable` despite being logically equivalent. fastfetch should subtract the size of ZFS's cache, similarly to other tools like htop. There's already ZFS-specific code in fastfetch for filesystem handling.

This can be implemented by reading `/proc/spl/kstat/zfs/arcstats` in addition to `/proc/meminfo` if it's present, and adding `size - c_min` to `MemAvailable`.

CarterLi:
Any additional information? Htop source? Is it required on FreeBSD and Solaris?

leo60228:
I don't know for sure but I assume not. The root cause of this is how the Linux kernel computes MemAvailable. The ZFS issue for it is here, but it doesn't show any path to being closed: https://github.com/openzfs/zfs/issues/10251

leo60228:
Here's the relevant code in [htop](https://github.com/htop-dev/htop/blob/0ba6e59cae04f77330594a90896e533197d693ff/linux/Platform.c#L430-L443) and [btop](https://github.com/aristocratos/btop/blob/9a97cb76a37efa8bed86e9ff8bfe74d2020d3ff3/src/linux/btop_collect.cpp#L1976-L2036).

CarterLi:
> The ZFS issue for it is here, but it doesn't show any path to being closed: [openzfs/zfs#10251](https://github.com/openzfs/zfs/issues/10251)

After reading the issue, especially someone running `sudo sync && echo 2 | sudo tee /proc/sys/vm/drop_caches` in crontab, I think it is a serious issue that should be fixed by openzfs themselves. 

I don't want to introduce special handling for a out-of-tree module in a common module, and doing it may introduce inconsistant behavior to FreeBSD and Solaris which also use zfs by default.

I'd rather not do this unless it can fix some issues like https://github.com/fastfetch-cli/fastfetch/issues/1988

leo60228:
[...]

PR #2243, opened 2026-03-25
(BSD) Exclude reclaimable ARC usage from used memory

biarder-dev:
## Summary

Add logic that calculates the memory usage except for reclaimable ARC pages on BSD

`ffDetectMemory` function on Linux systems looks like this.
It calculates memory usage except for *buffers* and *cached* memory.
```c
// src/memory/memory_linux.c:48
memAvailable = memFree + buffers + cached + sReclaimable - shmem;
```
BSD systems don't support those two pieces of information(it's just a *wired* in BSD).
Also the existing logic includes *wired* usage in memory usage! So the BSD's memory usage is calculated way more conservatively than on Linux. But the amount of wired memory is reclaimable by the kernel, because the kernel reclaims memory to the system when memory load is too high. [More Information](https://forums.freebsd.org/threads/question-about-wired-memory.58074/) Reclaimable ARC pages are the wired usage and also same as the cached memory. And so on, we can add more *pseudo cache* calculating codes in the function.

P.S I think [btop](https://github.com/aristocratos/btop/blob/main/src/freebsd/btop_collect.cpp) excludes wired memory usage for used memory section.

I tested with one firefox tab after the bootup on i3.
### Before
image

### After
image

The memory usage decreased from 20% (3.13 GiB) to 16% (2.54 GiB)!

Closes #N/A

## Changes

- Add `arcPages` to `pagesFree` in `ffDetectMemory` function.

## Checklist

### DIFF
--- a/src/detection/memory/memory_bsd.c
+++ b/src/detection/memory/memory_bsd.c
@@ -7,10 +7,17 @@ const char* ffDetectMemory(FFMemoryResult* ram)
     if (sysctl((int[]){ CTL_HW, HW_PHYSMEM }, 2, &ram->bytesTotal, &length, NULL, 0))
         return "Failed to read hw.physmem";
 
+    // calculates the reclaimable ARC pages.
+    int64_t arcSize = ffSysctlGetInt64("kstat.zfs.misc.arcstats.size", 0);
+    int64_t arcMin = ffSysctlGetInt64("vfs.zfs.arc_min", 0);
+    int64_t arcPages = (arcSize > arcMin) ? (arcSize - arcMin)
+        / instance.state.platform.sysinfo.pageSize : 0;
+
     // vm.stats.vm.* are int values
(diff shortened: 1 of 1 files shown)


CarterLi:
Related: #1995

I don't think the Linux implementation of memory module excludes ARC size of **zfs**, aka `/proc/spl/kstat/zfs/arcstats`, and I don't the `sReclaimable` size of Linux kernel will take the ARC size of an out-of-tree file system in to account. No?

> ```c
> // src/memory/memory_linux.c:48
> memAvailable = memFree + buffers + cached + sReclaimable - shmem;
> ```

It's a fallback. We use `MemAvaliable` which is calcuated directly by kernel. The exact algorithm is unknown.

> The memory usage decreased from 20% (3.13 GiB) to 16% (2.54 GiB)!

They are all estimation. Don't take them too seriously. In a sense the swap usage is much more useful. As long as you don't hit the swap, you are ok.

PR #2251, opened 2026-03-29
ci: pin third-party GitHub Actions to immutable SHAs

Elisha630:
Pins mutable `@master` GitHub Action refs in CI to immutable commit SHAs to reduce supply-chain risk. No functional behavior change intended.


flowerey:
Unneeded, plus, it would require manual updating.

CarterLi:
> Unneeded, plus, it would require manual updating.

I second this. If you worry about the security, you should build better fastfetch in source yourself.

CarterLi:
Done in #2492

PR #2253, opened 2026-03-30
fix(localip): use assigned netlink port id on Linux

### LINKED ISSUE #2252: [BUG] LocalIP: IPv6 default route detection fails due to netlink socket port ID mismatch
### Read the FAQ first

- [x] I have checked the FAQ but the issue is not covered

### Description

IPv6 default route detection silently fails when IPv4 default route detection runs first. Both `ffNetifGetDefaultRouteImplV4` and `ffNetifGetDefaultRouteImplV6` in `src/common/impl/netif_linux.c` filter netlink responses using `instance.state.platform.pid`, but the kernel only assigns the process PID as the netlink port ID (`nl_pid`) to the **first** socket bound with `nl_pid = 0`. Subsequent sockets receive a different kernel-assigned port. Since V4 runs first, it claims the PID port — then V6's response filter (`nlh->nlmsg_pid != pid`) drops every response because they arrive on a different port.

### Root cause

When binding a netlink socket with `nl_pid = 0`, the kernel tries `getpid()` first. If that's already taken (by the V4 socket), it assigns an arbitrary unique port. Both functions filter with:

```c
uint32_t pid = instance.state.platform.pid;
// ...
if (nlh->nlmsg_seq != 1 || nlh->nlmsg_pid != pid)
    continue;
```

The V6 socket's responses have a `nlmsg_pid` matching the kernel-assigned port, not the process PID, so they are all silently skipped.

### Suggested fix

After `bind()`, use `getsockname()` to retrieve the actual assigned port:

```c
struct sockaddr_nl bound_addr;
socklen_t bound_len = sizeof(bound_addr);
if (getsockname(sock_fd, (struct sockaddr*)&bound_addr, &bound_len) < 0)
    return false;
uint32_t pid = bound_addr.nl_pid;
```

This should be applied to both V4 and V6 functions for correctness, even though V4 happens to work today by being first. Happy to submit a PR if this approach looks good.

### Reproducer

Any system where V4 default route detection runs before V6 (the normal path) and an IPv6 default route exists:

```bash
ip -6 route show default   # confirms route exists
fastfetch --format json -s localip   # shows "ipv6": "" despite route existing
```

The issue is more visible on systems with a bridge or macvlan interface (e.g. `nm-bridge`), but affects any dual-stack system.

### Related

#1919 reported the same symptom but had a different root cause (multiple netlink packets). This is a separate issue.

### Version used

2.60.0

### Bug prevalence

Always

### Regression

No (the PID assumption has been present since the netlink implementation was added)

### Installation

Built from source

### System information

```
Arch Linux, kernel 6.18.18-1-lts, dual-stack with IPv6 default route
```

txhno:
## Summary
- read the kernel-assigned netlink port ID after `bind()`
- use that assigned port for both IPv4 and IPv6 route request/reply matching on Linux

Closes #2252

### DIFF
--- a/src/common/impl/netif_linux.c
+++ b/src/common/impl/netif_linux.c
@@ -7,6 +7,19 @@
 #include <linux/rtnetlink.h>
 #include <net/if.h>
 
+static bool ffNetifGetNetlinkPortId(int sock_fd, uint32_t* pid) {
+    struct sockaddr_nl addr = {};
+    socklen_t addrLen = sizeof(addr);
+    if (getsockname(sock_fd, (struct sockaddr*) &addr, &addrLen) < 0) {
+        FF_DEBUG("Failed to query netlink socket address: %s", strerror(errno));
+        return false;
+    }
+
+    *pid = addr.nl_pid;
+    FF_DEBUG("Netlink port ID: %u", *pid);
+    return true;
+}
+
 bool ffNetifGetDefaultRouteImplV4(FFNetifDefaultRouteResult* result) {
     FF_DEBUG("Starting IPv4 default route detection");
 
@@ -16,9 +29,6 @@ bool ffNetifGetDefaultRouteImplV4(FFNetifDefaultRouteResult* result) {
         return false;
     }
     FF_DEBUG("Created netlink socket: fd=%d", sock_fd);
-
-    uint32_t pid = instance.state.platform.pid;
-    FF_DEBUG("Process PID: %u", pid);
 
     // Bind socket
     struct sockaddr_nl addr = {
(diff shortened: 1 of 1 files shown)


src/common/impl/netif_linux.c
```diff
@@ -7,6 +7,19 @@
 #include <linux/rtnetlink.h>
 #include <net/if.h>
 
+static bool ffNetifGetNetlinkPortId(int sock_fd, uint32_t* pid) {
+    struct sockaddr_nl addr = {};
+    socklen_t addrLen = sizeof(addr);
+    if (getsockname(sock_fd, (struct sockaddr*) &addr, &addrLen) < 0) {
+        FF_DEBUG("Failed to query netlink socket address: %s", strerror(errno));
+        return false;
+    }
```
CarterLi, line 19: If getsockname fails for whatever reason, just set `*pid = instance.state.platform.pid`

PR #2288, opened 2026-04-25
fix small typo from versin to version

Or-simen:
## Summary
first time contributing to open source so i just fixed a typo 

Closes #

## Changes
changed a comment that said versin instead of version
 

## Checklist
no need to test this change

### DIFF
--- a/src/3rdparty/display-library/adl_structures.h
+++ b/src/3rdparty/display-library/adl_structures.h
@@ -39,55 +39,54 @@
 /// or fetch various settings upon the user's request.
 /// \nosubgrouping
 ////////////////////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


Vishrut2403 (commented):
Looks like you made changes in formatting like moving '{' onto the same line and the whitespace changes which makes the diff so large. You might want to revert those changes and keep only the typo fix

CarterLi:
Thanks for your contribution.

However, as the file path suggests, `src/3rdparty/display-library/adl_structures.h` is a 3rd-party file copied from <https://github.com/GPUOpen-LibrariesAndSDKs/display-library>. We don't make changes to these files unless absolutely necessary. You can make a PR to the original repo, and then fastfetch will update it.

PR #2299, opened 2026-05-03
add dpkg detection to macOS

evrett-k:
## Summary
Add dpkg package detection for macOS via Procursus bootstrap.

## Related issue
Closes #

## Changes
- `src/detection/packages/packages_apple.c`: Added `getDpkgPackages()` to detect Procursus dpkg packages

## Checklist
- [ ✓ ] I have tested my changes locally.

image

### DIFF
--- a/src/detection/packages/packages_apple.c
+++ b/src/detection/packages/packages_apple.c
@@ -6,11 +6,9 @@
 
 static void countBrewPackages(FFstrbuf* baseDir, FFPackagesResult* result) {
     uint32_t baseDirLength = baseDir->length;
-
     ffStrbufAppendS(baseDir, "/Caskroom");
     result->brewCask += ffPackagesGetNumElements(baseDir->chars, true);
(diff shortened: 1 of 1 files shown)


CarterLi:
What is Procursus bootstrap and why is it useful/important to macOS users?

abit-foggy:
> What is Procursus bootstrap and why is it useful/important to macOS users?

[Procursus](https://github.com/ProcursusTeam/Procursus) is a bootstrap typically used to install dpkg/apt on rootless jailbroken iOS and other related OSes (ex. tvOS). There is a more unknown version of this meant for Big Sur but it runs on the latest version of macOS. 

For your second statement it is mostly just a gimmick and has no real use as most of the binaries from most repos need to be recompiled for Mach-O

CarterLi:
> For your second statement it is mostly just a gimmick and has no real use as most of the binaries from most repos need to be recompiled for Mach-O

Then I don't think I want to accept this PR.

abit-foggy:
Alrighty then

PR #2314, opened 2026-05-09
Wallpaper (Haiku): Add support

mmuman:
## Summary

Add support for Haiku wallpaper. We check which is used for the current workspace to match on screenshots.

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1240,7 +1240,7 @@ elseif(Haiku)
         src/detection/tpm/tpm_nosupport.c
         src/detection/uptime/uptime_haiku.c
         src/detection/users/users_linux.c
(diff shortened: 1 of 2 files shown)


CarterLi:
Nice.

Off topic: Do you know if possible to get the SMBIOS entry address (`SMBIOSv2_ROOT_POINTER` and `SMBIOSv3_ROOT_POINTER`) in user space? A lot of modules (Host, BIOS, Board, and so on) use it.

https://cgit.haiku-os.org/haiku/commit/?id=6ce67a533615dde47dc598abc960cbf867d63ffb

mmuman:
I don't think we have an official way yet, IIRC for DMIdecode I just used `/dev/misc/mem` directly.

PR #2340, opened 2026-05-22
feat: add SDKMAN! package count support

### LINKED ISSUE #2027: [FEAT] Support SDKMAN! Package Count

gathddu:
## Summary

Adds support for counting installed SDKMAN! candidates in the Package module.

## Related issue

Closes #2027

## Changes

- Added `iunt32_t sdkman` field to `FFPackagesResult` in `packages.h`
- Added `FF_PACKAGES_FLAG_SDKMAN_BIT` flag to `option.h`
- Added detection logic in `packages_linux.c` that reads `$SDKMAN_CANDIDATES_DIR` and counts non-empty subdirectories
- Updated `packages.c` to print and export the sdkman count

## Checklist

- [X] I have tested my changes locally.

### DIFF
--- a/src/detection/packages/packages.h
+++ b/src/detection/packages/packages.h
@@ -43,6 +43,7 @@ typedef struct FFPackagesResult {
     uint32_t rpm;
     uint32_t scoopGlobal;
     uint32_t scoopUser;
+    uint32_t sdkman;
     uint32_t snap;
     uint32_t soar;
     uint32_t sorcery;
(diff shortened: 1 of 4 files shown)


CarterLi:
Duplicate: https://github.com/fastfetch-cli/fastfetch/pull/2033

PR #2358, opened 2026-05-30
Haiku port: more modules

mmuman:
## Summary

Adds bootmgr, brightness and wmtheme module implementations for Haiku.

## Checklist

### DIFF
--- a/src/detection/brightness/brightness_haiku.cpp
+++ b/src/detection/brightness/brightness_haiku.cpp
@@ -1,0 +1,46 @@
+extern "C" {
+#include "brightness.h"
+#include "common/strutil.h"
+}
+
+#include <Application.h>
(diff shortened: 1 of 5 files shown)


src/detection/brightness/brightness_haiku.cpp
```diff
@@ -0,0 +1,46 @@
...
+    // We need a valid be_app to query the app_server here.
+    BApplication app("application/x-vnd.fastfetch-cli-fastfetch");
+    BScreen screen{}; // default screen is the main one
+
+    do {
+        if (!screen.IsValid()) {
+            continue;
+        }
+
+        float value = std::nanf("0");
```
CarterLi: Please don't use NaN because it won't work with `-ffast-math`

PR #2390, opened 2026-06-09
detection/de: add GaudyStack Wayland compositor/DE

SuperChux:


CarterLi:
What is gaudystack? Your personal project? https://github.com/SuperChux/gaudystack is 404

I think it's too early to support it in fastfetch as the source repo doesn't even exist.

SuperChux:
Exists, private.

PR #2406, opened 2026-06-20
Packages (Linux): Improve emerge detection speed

keremgokcek:
## Summary

Makes emerge package count detection faster.

## Changes

Instead of listing every package's directory in every category to check if `SIZE` file exist, now it just counts the number of directories in every category folder. This decreases the number of directory listing operations by the number of packages. It's especially noticeable on cold boots.

Documentation of directory: https://wiki.gentoo.org/wiki//var/db/pkg

i7-5500U + SSD with 889 emerge and 11 flatpak user packages installed:
On cold boot: 350 ms -> 30 ms
On regular usage: 3 ms -> 0.9 ms

C2D T7200 + HDD with 438 emerge packages installed:
On cold boot: 2480 ms -> 1000 ms
On regular usage: 1.9 ms -> 0.8 ms

## Checklist

### DIFF
--- a/src/detection/packages/packages_linux.c
+++ b/src/detection/packages/packages_linux.c
@@ -458,6 +458,35 @@ static uint32_t getPacmanPackages(FFstrbuf* baseDir) {
     return getNumElements(baseDir, dbPath.chars, true);
 }
 
+static uint32_t getEmergePackagesImpl(FFstrbuf* baseDir)
+{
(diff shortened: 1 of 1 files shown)


CarterLi:
The old code was introduced in #131, 4 years ago

@ceamac Can you recall why did you choose to check if `SIZE` exists?

CarterLi:
This PR looks good to me (never used gentoo though). If @ceamac has comments, feel free to raise another issue.

PR #2440, opened 2026-07-05
Packages (pkgsrc): add dynamic prefix discovery

Suyun114:
## Summary

Previously pkgsrc detection hardcoded `/usr/pkg/pkgdb`, which missed installations at non-standard prefixes (e.g. `/opt/pkg`) and database directory layouts (e.g. `.pkgdb`).

Now `ffPackagesGetPkgsrc` finds the `pkgin` binary via `PATH`, derives the prefix, reads `<prefix>/etc/mk.conf` for `PKG_DBDIR`, and falls back to `<prefix>/pkgdb`. Analogous to `ffPackagesGetNix`, the function appends to `baseDir` for Bedrock Linux compatibility.

Support added for Linux, NetBSD, SunOS, and macOS.

## Changes

- New `src/detection/packages/packages_pkgsrc.c` — shared helper
- `packages_linux.c` / `packages_apple.c` / `packages_nbsd.c` / `packages_sunos.c` — all four now call `ffPackagesGetPkgsrc`
- `packages.h` — declaration guarded
- `CMakeLists.txt` — includes the new file

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -617,6 +617,7 @@ if(LINUX)
         src/detection/os/os_linux.c
         src/detection/packages/packages_linux.c
         src/detection/packages/packages_nix.c
+        src/detection/packages/packages_pkgsrc.c
         src/detection/poweradapter/poweradapter_linux.c
         src/detection/processes/processes_linux.c
         src/detection/gtk_qt/qt.c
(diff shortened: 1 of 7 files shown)


CarterLi:
Cleaned a bit.

I'm pretty sure that this PR will break bedrock. As this PR searches `pkgin` in $PATH, if `pkgin` is installed in one of bedrock strat, it will prepend the bedrock strat prefix twice. For example:

* baseDir: /bedrock/strata/distro
* pkgin path found in $PATH: /bedrock/strata/distro/usr/bin/pkgin
* ffStrbufAppendF(baseDir, "%s/pkgdb", pkginPath.chars): /bedrock/strata/distro/bedrock/strata/distro/usr/pkgdb

In addition, how bedrock strat handles `PKG_DBDIR` is still unknown.

Unless this is properly tested in bedrock, I don't think I want to merge it.

PR #2448, opened 2026-07-12
TerminalFont (Ghostty): parse config files statically instead of exec'ing `ghostty +show-config`

### LINKED ISSUE #2122: [BUG] ghostty detection is slow (not fast)
### Read the FAQ first

- [x] I have checked the FAQ but the issue is not covered

### Description

I was running `fastfetch` on terminal startup (e.g. in shell startup file) and noticed it seemed a bit slow, and I've narrowed it down (using `fastfetch --stat`) to the terminal and terminal font modules.

Expected: detecting terminal and its font is similar speed to other modules (e.g. less than 10ms).
Observed: it takes over 100ms. 

Example:
Image

Steps to reproduce:

1. Use ghostty terminal.
2. Run `fastfetch --stat`
3. Optionally, for debugging, use the simplified config (in ~/.config/fastfetch/config.jsonc) (attached below).
4. Observer timings.

### Version used

fastfetch 2.56.1 (x86_64)

### Bug prevalence

Always

### Regression

Not sure

### Installation

Package manager

### Package manager

brew

### Screenshots

_No response_

### Configuration

```jsonc
{
  "$schema": "https://github.com/fastfetch-cli/fastfetch/raw/master/doc/json_schema.json",
  "display": {
    "stat": false,
    "pipe": false,
    "showErrors": false,
    "disableLinewrap": true,
    "hideCursor": false,
    "separator": ": ",
    "color": {
      "keys": "",
      "title": "",
      "output": "",
      "separator": ""
    },
    "brightColor": true,
    "duration": {
      "abbreviation": false,
      "spaceBeforeUnit": "default"
    },
    "size": {
      "maxPrefix": "YB",
      "binaryPrefix": "iec",
      "ndigits": 2,
      "spaceBeforeUnit": "default"
    },
    "temp": {
      "unit": "D",
      "ndigits": 1,
      "color": {
        "green": "32",
        "yellow": "93",
        "red": "91"
      },
      "spaceBeforeUnit": "default"
    },
    "percent": {
      "type": [
        "num",
        "num-color"
      ],
      "ndigits": 0,
      "color": {
        "green": "32",
        "yellow": "93",
        "red": "91"
      },
      "spaceBeforeUnit": "default",
      "width": 0
    },
    "bar": {
      "char": {
        "elapsed": "■",
        "total": "-"
      },
      "border": {
        "left": "[ ",
        "right": " ]",
        "leftElapsed": "",
        "rightElapsed": ""
      },
      "color": {
        "elapsed": "auto",
        "total": "97",
        "border": "97"
      },
      "width": 10
    },
    "fraction": {
      "ndigits": 2
    },
    "noBuffer": false,
    "key": {
      "width": 0,
      "type": "string",
      "paddingLeft": 0
    },
    "freq": {
      "ndigits": 2,
      "spaceBeforeUnit": "default"
    },
    "constants": []
  },
  "general": {
    "thread": true,
    "processingTimeout": 5000,
    "detectVersion": true
  },
  "modules": [
    "break",
    {
      "type": "terminal",
      "keyIcon": ""
    },
    {
      "type": "terminalfont",
      "keyIcon": ""
    },
    "break",
    {
      "type": "colors",
      "key": " ",
      "keyIcon": "",
      "symbol": "block",
      "paddingLeft": 0,
      "block": {
        "width": 3,
        "range": [
          0,
          15
        ]
      }
    }
  ]
}
```

### System information

Probably unrelevant, seems issue stems from exec'ing ghostty (based on [this part of the code](https://github.com/fastfetch-cli/fastfetch/blob/ad7e23f64c2c857ca388ebe0bc6003e6ae201704/src/detection/terminalfont/terminalfont.c#L76)).

### Features built-in

```text
threads
vulkan
imagemagick7
chafa
zlib
sqlite3
libzfs
System yyjson
```

clicelee:
## Summary

`detectGhostty()` currently execs `ghostty +show-config`, which spawns the full ghostty binary — 42–262ms per run on an M-series Mac (2s+ on a cold first exec), making the terminalfont module by far the slowest one. This PR returns to static config file parsing, but now follows `config-file` directives, matching ghostty's documented/implemented semantics (verified against `loadRecursiveFiles()` in ghostty's `Config.zig`).

**Timings** (real Ghostty 1.3.1 window, `--stat`, macOS; the exec cost varies a lot run-to-run):

| | terminalfont module |
|---|---|
| before | 42–262 ms (2s+ on a cold first exec) |
| after | 0.12–0.40 ms |
| after, worst case (16 chained `config-file` includes) | 0.8–2 ms |

Closes #2122

## Changes

- `config-file` includes are queued and parsed **after** all root files, in discovery order (breadth-first, same as ghostty)
- Relative include paths resolve against the directory of the including file
- `?` optional prefix supported; missing files are skipped
- Each unique path loads once (cycle prevention, same as ghostty), plus a 16-file cap as a runaway safety net
- Root load order fixed to XDG → Application Support so later-loaded AppSupport values win, matching ghostty
- Removed the `+show-config` exec path entirely
- Fixed a latent bug where `font-size` was parsed and then overwritten with an empty buffer in the static (Muxy) path

#### The run captured below hit 261.9ms on the exec path; detection output is identical before/after.

Screenshot 2026-07-12 at 12 09 29 AM

#### Worst case with 16 chained `config-file` includes — still under 1ms, same detection result (`Fira Code (13pt)`) from both implementations:

Screenshot 2026-07-12 at 12 22 02 AM

## Checklist

### DIFF
--- a/src/detection/terminalfont/terminalfont.c
+++ b/src/detection/terminalfont/terminalfont.c
@@ -3,7 +3,6 @@
 #include "common/properties.h"
 #include "common/processing.h"
 #include "common/debug.h"
-#include "common/strutil.h"
 #include "detection/terminalshell/terminalshell.h"
 
 static void detectAlacritty(FFTerminalFontResult* terminalFont) {
@@ -67,14 +66,17 @@ static void detectAlacritty(FFTerminalFontResult* terminalFont) {
     ffFontInitMoveValues(&terminalFont->font, &fontFamily, &fontSize, &fontStyle);
 }
 
-static bool parseGhosttyConfig(FFstrbuf* path, FFstrbuf* fontName, FFstrbuf* fontNameFallback, FFstrbuf* fontSize) {
+// Maximum number of `config-file` directives to follow, guarding against runaway includes
+#define FF_GHOSTTY_MAX_CONFIG_FILES 16
+
+static void parseGhosttyConfig(const FFstrbuf* path, FFstrbuf* fontName, FFstrbuf* fontNameFallback, FFstrbuf* fontSize, FFlist* configFiles /* list of FFstrbuf */) {
     FF_DEBUG("parsing config: %s", path->chars);
 
     FF_STRBUF_AUTO_DESTROY buffer = ffStrbufCreate();
     FF_STRBUF_AUTO_DESTROY temp = ffStrbufCreate();
     if (!ffAppendFileBuffer(path->chars, &buffer)) {
         FF_DEBUG("cannot read config: %s", path->chars);
-        return false;
+        return;
     }
 
     char* line = nullptr;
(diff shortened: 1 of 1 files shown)


CarterLi:
LGTM. Thanks.

Issue #808, opened 2024-04-23
[FEAT] glibc version compatibility & statically linked builds


hykilpikonna:
## Wanted features:

Add some statically linked builds on github release.

## Motivation:

The github release for linux amd64 is not runnable on older versions of ubuntu 20 LTS.

![image](https://github.com/fastfetch-cli/fastfetch/assets/22280294/c35a0c2d-3d2c-4591-b5a1-64b3eacd2733)

![image](https://github.com/fastfetch-cli/fastfetch/assets/22280294/838509b6-2150-4071-8906-d4e26c4cee8e)

CarterLi:
If you don't trust me, you may try it yourself.

hykilpikonna:
> It won't work even for MUSL build. Otherwise it has been done years ago.

Okay, that's fair. Would it be possible to change the build target to lower glibc versions? (e.g. 2.12)? Afaik the latest glibc should be compatible with any version after 2.0.

CarterLi:
Fastfetch uses Github Action to build binaries, which supports only ubuntu 20.04 and newer.

The binaries released by fastfetch were built with 22.04. If you need old glibc version support, you may download the artifacts named [fastfetch-linux-old-amd64](https://github.com/fastfetch-cli/fastfetch/actions/runs/8796585816/artifacts/1438469592)

hykilpikonna:
> The binaries released by fastfetch were built with 22.04. If you need old glibc version support, you may download the artifacts named [fastfetch-linux-old-amd64](https://github.com/fastfetch-cli/fastfetch/actions/runs/8796585816/artifacts/1438469592)

Thanks! Do you want to maybe publish the linux-old build on GitHub releases as well? Or change the default to Ubuntu 20.04? It's a little bit strange to abandon support for 20.04 LTS since it's still being supported and hasn't reached EOL yet.

Azathothas:
> > The github release for linux amd64 is not runnable on older versions of ubuntu 20 LTS.
> 
> It was documented in the [README](https://github.com/fastfetch-cli/fastfetch?tab=readme-ov-file#linux)
> 
> > Add some statically linked builds on github release
> 
> It won't work even for MUSL build. Otherwise it has been done years ago.

Actually, it is possible to build statically linked musl binaries by changing a few cmake flags.
This is how I do it: 
- https://github.com/Azathothas/Toolpacks/blob/main/.github/scripts/x86_64_Linux/bins/fastfetch.sh
[...]

PR #2492, opened 2026-08-05
CI: hardens the workflows (SHA pins, scoped secrets and token permissions)

Totara-thib:
Drive-by CI hardening in three commits, one logical change each, found with the Plumber CLI (https://github.com/getplumber/plumber) and verified on dev.

**Commit 1 pins every action to its commit sha**, versions kept as comments. A tag like `@v1` or a branch like `@master` is a movable pointer: whoever controls the action, or anyone who compromises it, can re-point it and your next run executes their code. The spots that matter here: `build-release.yml` creates the GitHub release with the binaries users download, and `build-windows-hosts.yml` holds the SignPath token, so a re-pointed tag in those paths could ship a tampered or wrongly signed release. Same pattern as the tj-actions/changed-files incident (CVE-2025-30066). You already pin docker/setup-qemu-action by sha and you bumped action majors by hand before (#808), so this just finishes the job: every sha was resolved from the upstream repo and cross checked against its release tag, and the three `@master` refs (cross-platform-actions, setup-alpine, get-latest-release) are pinned to their latest release tag. The pins stay maintainable with a dependabot config for the `github-actions` ecosystem (the repo has none today, happy to add one here if you want); dependabot and renovate both bump sha pins and keep the version comment in sync.

**Commit 2 drops `secrets: inherit` from all 17 reusable workflow calls.** The only secret any called workflow reads is `SIGNPATH_API_TOKEN` in `build-windows-hosts.yml`, so that one is declared and passed explicitly and nothing else gets the secret set. A compromised step in one of the BSD build VMs now has no signing token in reach. The secret is optional in the called workflow, so fork and PR runs behave exactly as before (the signing step is already guarded to upstream pushes).

**Commit 3 scopes the GITHUB_TOKEN per workflow**, derived from what each job actually does with it: every build workflow declares `contents: read`; `build-linux-hosts.yml` keeps `security-events: write` for its CodeQL upload while the other callers lose that grant (they never upload scanning results); `build-windows-hosts.yml` gets `actions: read`, which the SignPath action documents needing to read job details and fetch the unsigned artifact; `build-release.yml` declares the `contents: write` its caller already grants. Runs that execute PR code now hold a read-only token.

Two heads up:

- If the org uses an Actions allowlist in settings, patterns written against tags (like `owner/action@v1`) stop matching once refs are shas and workflows refuse to start. Entries need to be `owner/action@*` in that case.
- If you would rather not have a third-party action fetching the latest release tag in the release path at all, `pozetroninc/github-action-get-latest-release` can be replaced with a two-line `gh api` step; happy to do that here, kept the change minimal for now.

I will also open a PR which adds the tool to CI so this does not quietly drift back; that one is a bonus, this PR stands on its own.

Not a logo PR, no related issue.

## Changes

- Pin all 51 action references across the 19 workflows to full commit SHAs with version comments
- Replace `secrets: inherit` with an explicit, declared `SIGNPATH_API_TOKEN` pass to the one workflow that uses it
- Declare least-privilege `GITHUB_TOKEN` permissions on every workflow, derived from actual token use

## Screenshots

No visual changes.

## Checklist

### DIFF
--- a/.github/workflows/build-dragonfly-amd64.yml
+++ b/.github/workflows/build-dragonfly-amd64.yml
@@ -12,10 +12,10 @@ jobs:
 
     steps:
       - name: checkout repository
-        uses: actions/checkout@v7
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
       - name: Start VM
-        uses: cross-platform-actions/action@master
+        uses: cross-platform-actions/action@5ea7e8e4677bd726033a10b094ba1c5762b15dee # v1.3.0
         with:
           operating_system: dragonflybsd
           version: '6.4.2'
@@ -57,7 +57,7 @@ jobs:
         run: ctest --output-on-failure
 
       - name: upload artifacts
-        uses: actions/upload-artifact@v7
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
           name: fastfetch-dragonfly-amd64
           path: ./fastfetch-*.*
--- a/.github/workflows/build-dragonfly-amd64.yml
+++ b/.github/workflows/build-dragonfly-amd64.yml
@@ -3,6 +3,9 @@ name: Reusable DragonFly amd64
 on:
   workflow_call:
 
+permissions:
+  contents: read
+
 jobs:
   build:
     runs-on: ubuntu-latest
--- a/.github/workflows/build-freebsd-amd64.yml
+++ b/.github/workflows/build-freebsd-amd64.yml
@@ -3,6 +3,9 @@ name: Reusable FreeBSD amd64
 on:
   workflow_call:
 
+permissions:
+  contents: read
+
 jobs:
   build:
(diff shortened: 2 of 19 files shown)


flowerey:
So unneeded and not even useful. Why is this even merged?

Totara-thib:
> So unneeded and not even useful. Why is this even merged?

Fair question. A concrete case: March 2025, tj-actions/changed-files (CVE-2025-30066). The attacker re-pointed the action's existing version tags at code that dumped CI secrets into public build logs, and over 20k repos ran it on their next build without changing a line. Repos pinned by commit SHA were untouched: the tags moved, their refs did not.

That pattern matters here because this CI holds real credentials. The Windows job carries the SignPath token that signs the release binaries, and the release workflow publishes what thousands of people download. One re-pointed tag on those paths and you get a tampered fastfetch, signed and shipped under the project's name. The secrets change is the same idea: before, all 17 build jobs received every repo secret; now only the signing job sees the signing token, so a compromised step in one of the BSD build VMs has nothing to steal.

Runtime cost is zero, the refs are just frozen, and dependabot keeps bumping them (it already did this morning)

PR #2498, opened 2026-08-06
CMake (Hurd): Define O_PATH to O_NORW.

yelninei:
This is a noop because O_RW is 0 but O_NORW is the correct constant to open a file without read/write permissions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1555,7 +1555,7 @@ elseif(Haiku)
     target_compile_definitions(libfastfetch PUBLIC _GNU_SOURCE)
 elseif(GNU)
     # On Hurd PATH_MAX is not defined. Set an arbitrary limit as workaround.
-    target_compile_definitions(libfastfetch PUBLIC _GNU_SOURCE PATH_MAX=4096 O_PATH=0)
+    target_compile_definitions(libfastfetch PUBLIC _GNU_SOURCE PATH_MAX=4096 O_PATH=O_NORW)
 endif()
 
 if(APPLE)


CarterLi:
Do you mean `O_NORW is 0`?

yelninei:
Yes, but inlike `0` which I put there initially to make the missing O_PATH be ignored it is defined to roughly the same as O_PATH (and because read/write permissions are a bitmask on Hurd 0 is accidentally the right flag for that)

```c
/* In GNU, read and write are bits (unlike BSD).  */
#ifdef __USE_GNU
# define O_READ		O_RDONLY /* Open for reading.  */
# define O_WRITE	O_WRONLY /* Open for writing.  */
# define O_EXEC		0x0004	/* Open for execution.  */
# define O_NORW		0	/* Open without R/W access.  */
#endif
/* POSIX.1 standard names.  */
#define	O_RDONLY	0x0001	/* Open read-only.  */
#define	O_WRONLY	0x0002	/* Open write-only.  */
#define	O_RDWR		(O_RDONLY|O_WRONLY) /* Open for reading and writing. */
#define	O_ACCMODE	O_RDWR	/* Mask for file access modes.  */
```

I found out about O_NORW while browsing the debian package fir procps https://salsa.debian.org/debian/procps/-/blob/master/debian/patches/no-opath

CarterLi:
> 0 is accidentally the right flag for that

It's not *accidentally*. `| 0` is a noop. `0` must be no read/write as long as you don't write `0 | O_RDONLY` or `0 | O_WRONLY`.

I still prefer `0` because `O_NORW` is not standard and I never saw it before.

yelninei:
> I still prefer 0 because O_NORW is not standard

Neither is O_PATH which is linux specific (it seems to also exists in the recent freebsd versions for linux compatibilty)

> 0 must be no read/write

That is wrong in general but true on Hurd where `O_RDWR = (O_RDONLY|O_WRONLY)`.

On linux O_RDONLY is 0 so in the general open flags of 0 are not no r/w. 

https://sourceware.org/glibc/manual/latest/html_node/Access-Modes.html

```
Macro: int O_PATH

    Obtain a file descriptor for the file, but do not open the file for reading or writing. Permission checks for the file itself are skipped when the file is opened (but permission to access the directory that contains it is still needed), and permissions are checked when the descriptor is used later on.

    For example, such descriptors can be used with the fexecve function (see Executing a File). Other applications involve the ‘*at’ function variants, along with the AT_EMPTY_PATH flag. See Descriptor-Relative Access.

    This access mode is specific to Linux. On GNU/Hurd systems, it is possible to use O_EXEC explicitly, or specify no access modes at all (see below). 
``` 

I put it there to make `x | O_PATH` a noop to ignore that O_PATH is not defined. But (accidentaly) by the way the openflags are implemented this is also how O_NORW works but with a guaranteed permisson semantic instead.

PR #2514, opened 2026-08-13
Doc: restore Star History chart in README.md

### LINKED ISSUE #2508: Star History Broken in Readme
The star history in the readme file is broken. Apparently, you can only see the chart if you're a contributor to the project. the star-history blog says something about a particular token is now being required.

https://www.star-history.com/blog/github-stargazer-api-restriction/

Here's what it looks like to non-contributors:

Image

OctoBored:
The Star History chart was removed in 8df0a1c9 (Fixes #2508) because the previous chart was broken due to GitHub stargazer API restrictions. This change restores it at the same location using a working data source that requires no API token, so the chart renders correctly again.


CarterLi:
> using a working data source that requires no API token

This is interesting. Thanks.

PR #2565, opened 2026-09-04
CMake: bump min version to 3.21

Guiorgy:
C_STANDARD C17 and C23 were only introduced in CMake 3.21

CMake older than `3.21` give the `C_STANDARD is set to invalid value '23'` error.

## Checklist

Also tested the workflow in my fork:

workflow creates *-i686 packages

## Notes

Between versions `3.12` and `3.21` the [CMAKE_SYSTEM_PROCESSOR](https://cmake.org/cmake/help/v3.21/variable/CMAKE_SYSTEM_PROCESSOR.html) variable added:

> When cross-compiling, a [CMAKE_TOOLCHAIN_FILE](https://cmake.org/cmake/help/v3.21/variable/CMAKE_TOOLCHAIN_FILE.html#variable:CMAKE_TOOLCHAIN_FILE) should set the `CMAKE_SYSTEM_PROCESSOR` variable to match target architecture that it specifies (via [CMAKE_<LANG>_COMPILER](https://cmake.org/cmake/help/v3.21/variable/CMAKE_LANG_COMPILER.html#variable:CMAKE_%3CLANG%3E_COMPILER) and perhaps [CMAKE_<LANG>_COMPILER_TARGET](https://cmake.org/cmake/help/v3.21/variable/CMAKE_LANG_COMPILER_TARGET.html#variable:CMAKE_%3CLANG%3E_COMPILER_TARGET)).

So by that text it seems that `CMAKE_TOOLCHAIN_FILE` is required, however, the page for [CMAKE_SYSTEM_NAME](https://cmake.org/cmake/help/v3.21/variable/CMAKE_SYSTEM_NAME.html) specifies:

> `CMAKE_SYSTEM_NAME` may be set explicitly when first configuring a new build tree in order to enable [cross compiling](https://cmake.org/cmake/help/v3.21/manual/cmake-toolchains.7.html#cross-compiling-toolchain). In this case the [CMAKE_SYSTEM_VERSION](https://cmake.org/cmake/help/v3.21/variable/CMAKE_SYSTEM_VERSION.html#variable:CMAKE_SYSTEM_VERSION) variable must also be set explicitly.

Which I take it as using `CMAKE_SYSTEM_NAME` and `CMAKE_SYSTEM_VERSION` being enough to enable cross-compilation? I tested exactly this (added them to the workflow) and it seems to have resolved the issue, cross-compilation works and produces `*-i686` artifacts, and this is the least intrusive change. However, the text about `CMAKE_SYSTEM_NAME` enabling cross-compilation was present even in `3.12`, so I honestly don't know why it worked before without it, and why it works now with it. It may be better to also follow the instruction to use `CMAKE_TOOLCHAIN_FILE` just in case, though that will require creating the file. Let me know if you do want to do that, or leave it for when/if this becomes an issue again.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.12.0) # Don't change before we investigate why #2553 happens
+cmake_minimum_required(VERSION 3.21.0) # C_STANDARD C17 and C23
 
 project(fastfetch
     VERSION 2.68.1
@@ -9,9 +9,6 @@ project(fastfetch
 
 set(PROJECT_LICENSE "MIT license")
 
-if(DEFINED CMAKE_SYSTEM_PROCESSOR_OVERRIDE) # Used by github actions for i686 build
-    set(CMAKE_SYSTEM_PROCESSOR ${CMAKE_SYSTEM_PROCESSOR_OVERRIDE} CACHE INTERNAL "")
-endif()
 string(TOLOWER "${CMAKE_SYSTEM_PROCESSOR}" CMAKE_SYSTEM_PROCESSOR)
 if(CMAKE_SYSTEM_PROCESSOR STREQUAL "x86_64")
     set(CMAKE_SYSTEM_PROCESSOR "amd64")
--- a/.github/workflows/build-linux-i686.yml
+++ b/.github/workflows/build-linux-i686.yml
@@ -40,7 +40,7 @@ jobs:
         run: cmake --version
 
       - name: configure project
-        run: CC=gcc-13 CMAKE_PREFIX_PATH=/home/linuxbrew/.linuxbrew PKG_CONFIG_PATH=/home/linuxbrew/.linuxbrew/lib/pkgconfig:$PKG_CONFIG_PATH cmake -DCMAKE_C_FLAGS="-m32 -march=i686 -mtune=i686" -DCMAKE_SYSTEM_PROCESSOR_OVERRIDE=i686 -DCPACK_DEBIAN_PACKAGE_ARCHITECTURE=i386 -GNinja -DSET_TWEAK=Off -DBUILD_TESTS=On -DENABLE_EMBEDDED_PCIIDS=On -DENABLE_EMBEDDED_AMDGPUIDS=On -DCMAKE_INSTALL_PREFIX=/usr .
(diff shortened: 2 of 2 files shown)


CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.12.0) # Don't change before we investigate why #2553 happens
+cmake_minimum_required(VERSION 3.12.0) # C_STANDARD C17 and C23
```
CarterLi: You didn't?

Guiorgy: Facepalm, overwrote it in a forced push! I am away to a meeting, will redouble check everything tonight/tomorrow, rorry 😂

Guiorgy: I wanted to test the GitHub workflows (first time using them) for `3.12` and had forgotten to switch it back :D, anyway after the forced push `*-i686` artifacts are correctly generated

Guiorgy:
Also downloaded the build artifacts for i686 and run a docker container to test them just in case:

```console
> docker run --rm -it -v ./:/ff -w /ff --platform linux/386 debian:stable bash
# getconf LONG_BIT
32
# dpkg --print-architecture
i386
# apt update && apt install -y file
# apt install -y ./fastfetch-linux-i686.deb
# file $(command -v fastfetch)
/usr/bin/fastfetch: ELF 32-bit LSB pie executable, Intel i386, version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux.so.2, BuildID[sha1]=93addc9f72c2191d6e64a93b213c0191d4b88c5e, for GNU/Linux 3.2.0, with debug_info, not stripped
# fastfetch
        _,met$$$$$gg.          root@526cf1a019b1
     ,g$$$$$$$$$$$$$$$P.       -----------------
   ,g$$P""       """Y$$.".     OS: Debian GNU/Linux 13.6 (trixie) x86_64
  ,$$P'              `$$$.     Kernel: Linux 5.15.146.1-microsoft-standard-WSL2
',$$P       ,ggs.     `$$b:    Uptime: 8 mins
`d$$'     ,$P"'   .    $$$     Packages: 82 (dpkg)
 $$P      d$'     ,    $$P     Shell: bash 5.2.37
 $$:      $$.   -    ,d$$'     Terminal: xterm
 $$;      Y$b._   _,d$P'       Terminal Font: fixed (8.0pt)
 Y$$.    `.`"Y$$$$P"'          CPU: AMD Ryzen 9 5950X (6) @ 3.40 GHz
 `$$b      "-.__               GPU: Microsoft Basic Render Driver
  `Y$$b                        Memory: 776.52 MiB / 11.69 GiB (6%)
   `Y$$.                       Swap: 0 B / 4.00 GiB (0%)
     `$$b.                     Disk (/): 34.81 GiB / 1006.85 GiB (3%) - overlay
       `Y$$b.                  Disk (/ff): 137.05 GiB / 349.31 GiB (39%) - 9p
         `"Y$b._               Local IP (eth0): 172.17.0.2/16
             `""""             Locale: C
```

Edit: Interestingly `uname -m` prints `x86_64` inside the container, even though I run it with `--platform linux/386` and `dpkg --print-architecture` gives `i386`.

PR #2566, opened 2026-09-05
CPU: show heterogeneous RISC-V microarchitectures

RevySR:
## Summary

Show a per-microarchitecture CPU breakdown on heterogeneous RISC-V systems, even when a SoC name has already been detected.

For example, the A210 contains four C908 and four C920 cores. Both clusters report the same maximum frequency, so frequency-based grouping alone does not distinguish them.

N/A

## Changes

- Parse and group per-CPU `uarch` descriptions from `/proc/cpuinfo`.
- Preserve the SoC name and append counts for distinct microarchitectures.
- Keep existing behavior for homogeneous systems and incomplete CPU descriptions.
- Leave frequency reporting and the `coreTypes` interface unchanged.
- Add tests for heterogeneous, interleaved, homogeneous, and malformed CPU descriptions.

## Screenshots

Expected output:

image

Before:

```text
CPU: a210 (8) @ 1.90 GHz
```

After:

```text
CPU: a210 (4 x thead,c908 + 4 x thead,c920) (8) @ 1.90 GHz
```

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2130,6 +2130,8 @@ endif()
 ###################
 
 if (BUILD_TESTS)
+    add_executable(fastfetch-test-cpu-riscv tests/cpu-riscv.c)
+    target_link_libraries(fastfetch-test-cpu-riscv PRIVATE libfastfetch)
     add_executable(fastfetch-test-strbuf
         tests/strbuf.c
     )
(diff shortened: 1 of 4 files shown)


CarterLi:
I don't like this. I think this is against fastfetch's design and it looks like an AI slop to me (I'm sorry if it isn't)

PR #2572, opened 2026-09-08
CPU (Linux): adds SPARC support

shalseth:
## Summary

`/proc/cpuinfo` on SPARC has no `model name`, `model`, `cpu model`, `hardware` or `processor`
field, so the CPU module prints `Unknown`. The chip name is in the `cpu` field, and the clock
frequency is in `Cpu<n>ClkTck` — in Hz rather than MHz, and in hex on sparc64. SPARC has no
`cpufreq` sysfs, so that field is the only frequency source.

```diff
  # UltraSPARC T4-1 (sun4v)
- CPU: Unknown (64)
+ CPU: UltraSparc T4 (Niagara4) (64) @ 2.85 GHz

  # Sun Fire V240 (sun4u)
- CPU: Unknown (2)
+ CPU: TI UltraSparc IIIi (Jalapeno) (2) @ 1.50 GHz
```

## Related issue

`/proc/cpuinfo` for both machines is in #1529.

## Changes

- `parseCpuInfo()`: under `__sparc__`, read the name from `cpu` and the clock tick from `Cpu0ClkTck`.
- `detectCPUOthers()`: convert that from Hz to MHz. sparc64 prints it as `%016lx` and sparc32 as
  `%ld`; since a 32-bit userland can run on a 64-bit kernel, the base comes from the value's width
  rather than from the build's bitness.

Core counts already work and are unchanged (8/64 on the T4-1, 2/2 on the V240). sparc32 is untested.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# Unreleased
+
+Features:
+* Added CPU name and frequency detection support on SPARC. (CPU, Linux)
+
 # 2.68.1
 
 Changes:
--- a/src/detection/cpu/cpu_linux.c
+++ b/src/detection/cpu/cpu_linux.c
@@ -619,6 +619,9 @@ static const char* parseCpuInfo(
             (cpu->name.length == 0 && ffParsePropLine(line, "cpu :", &cpu->name)) ||
 #elif __sh__
             (cpu->name.length == 0 && ffParsePropLine(line, "cpu type :", &cpu->name)) ||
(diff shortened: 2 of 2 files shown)


CarterLi:
You should add a new entry `UNRELEASED` in CHANGELOG since 2.68.1 has been released.

Other changes LGTM.

shalseth:
Thanks, CHANGELOG updated.

PR #2585, opened 2026-09-14
test(strbuf): cover the length-bounded contract of ffStrbufMatchSeparatedNS

buyua9:
## What does this PR do?
Adds a focused test block for `ffStrbufMatchSeparatedNS`, the length-bounded variant behind `ffStrbufMatchSeparatedS`.

## Why
The Linux keyboard handler calls the NS variant with a fixed-size buffer length rather than a NUL-terminated string, but `tests/strbuf.c` only exercised the S wrapper — the bounded contract had zero coverage. The new block asserts twelve length-boundary cases: bounds cutting mid-token (match vs no-match), `compLength = 0`, a bound spanning exactly one token, no separator in bounds, an embedded NUL inside the bounds, and empty-substring matching, all against the current implementation semantics.

## Tested
`cmake -DFF_BUILD_TESTS=on` → `ffstrbuf_test` (block verified against the implementation's branch behavior; full suite runs in CI).


tests/strbuf.c
```diff
@@ -681,6 +681,26 @@ int main(void) {
...
+    {
+        // ffStrbufMatchSeparatedNS with explicit compLength: the length-bounded
+        // contract is load-bearing for the Linux keyboard handler, which passes
+        // a fixed-size buffer length rather than a NUL-terminated string.
+        ffStrbufSetStatic(&strbuf, "abc");
+        VERIFY(ffStrbufMatchSeparatedNS(&strbuf, 11, "abc:def:ghi", ':') == true);
+        VERIFY(ffStrbufMatchSeparatedNS(&strbuf, 3, "abc:def:ghi", ':') == true);
+        VERIFY(ffStrbufMatchSeparatedNS(&strbuf, 7, "abc:def:ghi", ':') == true); // "abc:def"
+        VERIFY(ffStrbufMatchSeparatedNS(&strbuf, 4, "abc:def:ghi", ':') == true); // "abc:"
+        VERIFY(ffStrbufMatchSeparatedNS(&strbuf, 5, "abc:def:ghi", ':') == false); // "abc:d"
```
CarterLi, line 693: This means if `"abc:d"` contains `"abc"`. This is obviously true.

PR #2597, opened 2026-09-20
Packages (Linux): detects rum overlay packages

### LINKED ISSUE #2511: [FEAT] support for listing rum package count
### Description

I would like to see the ability for fastfetch to count packages installed with rum. Rum is RakuOS's new rpm based package manager. Here's the thing, rum is able to separate "base" and "overlay". fastfetch should have the ability to count packages. Also, `rpm -qa` only lists base packages and not user packages on RakuOS.

### Motivation

Well, this isn't just for me and that's it. It is for an entirely new package manager (uses rpm package format) that is used for RakuOS which has an overlayfs mount on /usr so there is the atomic base where you can install any rpm on the overlay.

### Additional context

Using the command `rum list` will list all user installed packages. `rpm -qa` only lists base packages and when  I run `rpm -qa | count` and `rum list --base | count` the package count is only off by 6 packages (rpm -qa lists slightly more for some reason) while non of the packages installed by the user are listed unless `rum list` is used and when I type `rum list | count`  I have 471 packages installed.

Edit: It may be better to use `rpm --dbpath /var/lib/rakuos/rum-rpmdb -qa | wc -l` since it is faster. Maybe it can be listed as "rpm (base)" and "rpm (overlay)" for the lable kind of like how there is "flatpak (system)" and "flatpak (user)"

ddy314:
## Summary

Count packages in RakuOS rum's overlay RPM database using the existing cached SQLite lookup. Keep the count separate from the existing `rpm` field.

## Related issue

Closes #2511

## Changes

- Read `/var/lib/rakuos/rum-rpmdb/rpmdb.sqlite`, expose `rum` in default output, JSON, custom formats, and the disabled list.
- Place the format argument and help entry after `rpm` in alphabetical order, and update the schema and changelog.

## Validation

Built with GCC on Linux and passed all 16 CTest tests. Eight checks using an isolated target root and SQLite fixtures cover independent base/overlay counts, cache reuse and invalidation, disabled detection, named formats, default output, and missing/empty databases. The base commit reports only the three fixture RPM packages; the patched build reports three RPM plus five rum packages, totaling eight.

The normal host build also reports its existing pacman package count correctly. RakuOS itself was not booted for testing. Optional Vulkan and librpm support were disabled in the local build; SQLite support was enabled. The all-modules smoke run exceeded a 40-second limit, so only the Packages smoke and CTest results are claimed here.

## Screenshots

No GUI changes. Fixture output: `3 (rpm), 5 (rum)`.

## Checklist

AI assistance: Codex helped implement and test this change and draft this description.


src/modules/packages/packages.c
```diff
@@ -202,6 +203,7 @@ bool ffPrintPackages(FFPackagesOptions* options) {
                 FF_ARG(hpkgAll, "hpkg-all"),
                 FF_ARG(nixAll, "nix-all"),
                 FF_ARG(counts.all, "all"),
+                FF_ARG(counts.rum, "rum"),
```
CarterLi: Should be in alphabet order, before all `-all` entries.

src/modules/packages/packages.c
```diff
@@ -601,6 +606,7 @@ FFModuleBaseInfo ffPackagesModuleInfo = {
         { "Total number of all hpkg packages", "hpkg-all" },
         { "Total number of all nix packages", "nix-all" },
         { "Number of all packages", "all" },
+        { "Number of rum overlay packages", "rum" },
```
CarterLi: ditto

src/modules/packages/packages.c
ddy314: Moved rum immediately after rpm in both the format argument list and the help entries in 6a57e65. Rebuilt and reran all 16 CTest tests plus the eight isolated RPM/rum fixture checks; all passed.

